PR #7, opened 2015-05-25
Round percentages in comparison output to 1 decimal

lencioni:
These floats were pretty long, with way more significant figures than is
actually useful. Rounding it to 1 decimal place cleans up the output a
little.


trotzig:
Thanks, that's better. Those percentages are pretty unusable as is anyway. I think they might have made sense at some point in a Diffux world, but that was before we switched to diff-lcs.

lencioni:
In Diffux we are using them to render little charts that express the percentage when viewing all of the thumbnails.

PR #9, opened 2015-07-17
Add more instructions to the readme

lencioni:
It isn't obvious that you need to run `likadan review` after the likadan
has finished. Adding this information to the readme will make this tool
much easier for people to pick up.


trotzig:
Thanks @lencioni! I've been meaning to add this but kept forgetting...

PR #26, opened 2015-08-02
Set <body> min-height to 1px on runner document

### LINKED ISSUE #23: Invalid image size height 0
I noticed this error in the output of a Jenkins run (/job/brigade-likadan/524/console). Not sure if it causing actual problems, but I thought that likadan could probably at least give us a more instructive error message.

```
/var/lib/jenkins/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/chunky_png-1.3.4/lib/chunky_png/canvas/png_decoding.rb:76:in `from_datastream': Invalid image size, width: 320, height: 0 (ChunkyPNG::ExpectationFailed)
	from /var/lib/jenkins/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/chunky_png-1.3.4/lib/chunky_png/image.rb:74:in `from_datastream'
	from /var/lib/jenkins/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/chunky_png-1.3.4/lib/chunky_png/canvas/png_decoding.rb:53:in `from_file'
	from /var/lib/jenkins/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/likadan-0.0.13/lib/likadan_runner.rb:70:in `block in <top (required)>'
	from /var/lib/jenkins/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/likadan-0.0.13/lib/likadan_runner.rb:31:in `each'
	from /var/lib/jenkins/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/likadan-0.0.13/lib/likadan_runner.rb:31:in `<top (required)>'
	from /var/lib/jenkins/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/likadan-0.0.13/bin/likadan:13:in `require'
	from /var/lib/jenkins/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/likadan-0.0.13/bin/likadan:13:in `block in <top (required)>'
```

lencioni:
I was using likadan recently and came across an interesting error:

> Invalid image size, width: 320, height: 0

I have an example that renders a component that happens to take up 0
height. Likadan is just fine with this on the first pass, but on the
second pass when comparing a new version against a baseline, it
blows up with this error.

To prevent this from happening, I am setting a min-height of 1px on the
<body> element. If I understand how this works correctly, this will
cause the snapshot to be at least 1px tall, so it should never store a
0px tall image and run into this problem.

We may want to implement a more thorough fix for this issue at some
point, but this is simple and should at least help people out now.

Fixes #23

### DIFF
--- a/lib/views/index.erb
+++ b/lib/views/index.erb
@@ -17,7 +17,7 @@
         href="/resource?file=<%= ERB::Util.url_encode(stylesheet) %>">
     <% end %>
   </head>
-  <body style="background-color: #fff; margin: 0; pointer-events: none;">
+  <body style="background-color: #fff; margin: 0; min-height: 1px; pointer-events: none;">
     <script src="/likadan-runner.js"></script>
     <% @config['source_files'].each do |source_file| %>
       <script src="/resource?file=<%= ERB::Util.url_encode(source_file) %>"></script>


trotzig:
Thanks for the PR!

PR #30, opened 2015-08-03
Avoid creating multiple S3 buckets

lencioni:
We started seeing the following error when likadan runs in CI:

> You have attempted to create more buckets than allowed
> (S3::Error::TooManyBuckets)

I think we can avoid this by only creating one bucket instead of a new
bucket every time likadan runs. I simply moved the random bucket name to
be a random file name, and created a non-random bucket name.

I decided to switch from SecureRandom.hex to SecureRandom.uuid because
uuid seems more appropriate for this type of thing.

I considered making the filename for the diffs just be the MD5 hash of
the file, which would avoid us from uploading duplicate data if the two
files happened to be the same. However, I didn't think the little bit of
added complexity of finding if the file already exists was worth the
small gain of no duplicates at this time.

I wasn't exactly sure how to test this, so I'm not sure if it actually works.

### DIFF
--- a/lib/likadan_uploader.rb
+++ b/lib/likadan_uploader.rb
@@ -16,8 +16,10 @@ class LikadanUploader
 
     bucket = find_or_build_bucket
 
+    dir = SecureRandom.uuid
+
     diff_images = current_snapshots[:diffs].map do |diff|
-      image = bucket.objects.build("#{diff[:name]}_#{diff[:viewport]}.png")
+      image = bucket.objects.build("#{dir}/#{diff[:name]}_#{diff[:viewport]}.png")
       image.content = open(diff[:file])
       image.content_type = 'image/png'
       image.save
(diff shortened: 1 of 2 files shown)


lib/likadan_uploader.rb
```diff
@@ -36,4 +37,17 @@ def upload_diffs
     html.save
     html.url
   end
+
+  private
+
+  def find_or_build_bucket
+    bucket = service.buckets.find(BUCKET_NAME)
```
sds: Looks like you'll need to pass `service` as an argument to this method in order for this to work.

lencioni: doh! I'll push a new version soon.

lencioni:
I ended up tacking on another unrelated commit, since it might help us avoid some random failures of likadan in CI.

PR #46, opened 2015-10-18
Run screenshot and image comparison in separate threads

lencioni:
There is a clean and easy division of work here. The first thread runs
Firefox, rendering the examples, taking screenshots, cropping them, and
pushing that onto a queue. The second thread reads from the queue,
comparing the cropped screenshot to the baseline and generating diff
images. This allows us to take more advantage of systems that have
multiple CPUs, which will speed up this process a little.

In an unscientific test I ran on my laptop on the Brigade diffux test
suite using `time diffux_ci`, this brought a clean run where there were
no visual diffs down from 194s to 170s. This is a ~12% improvement and I
imagine there might be even more savings for runs that have diffs. Since
we run through this twice for every commit in CI, I would expect this to
bring our CI job runtimes down by maybe a minute.

We could probably explore taking this technique a step further by
increasing the number of threads to match the number of CPUs or some
configurable concurrency value and dividing up the work that way, but
this seems like a good enough start for now.

I am not very familiar with threads in Ruby, so it is likely that I have
made some mistakes here, but it seems to run on my laptop so I feel
pretty comfortable with this change.

### DIFF
--- a/lib/diffux_ci_runner.rb
+++ b/lib/diffux_ci_runner.rb
@@ -10,8 +10,6 @@ require 'diffux_ci_utils'
 require 'fileutils'
 require 'thread'
 
-END_OF_QUEUE = 1
-
 def resolve_viewports(example)
   configured_viewports = DiffuxCIUtils.config['viewports']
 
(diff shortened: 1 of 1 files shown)


lencioni:
I think the threading might not help at all under MRI, but the `:fast_rgb` commit will help everything.

lencioni:
I moved the `:fast_rgb` out of this PR and put it on master.

lencioni:
I think that we happen to be CPU bound, so if we want to see any real wins here we probably want to use process based concurrency via `fork`. More info: https://blog.engineyard.com/2010/concurrency-real-and-imagined-in-mri-threads

lencioni:
We went down the `fork` path and were unable to get any improvements. Closing this out.

PR #48, opened 2015-10-24
Shorten command from `diffux_ci` to `diffux`

### LINKED ISSUE #43: Shorten command from diffux_ci to just diffux?
It is pretty unusual for a command to have an underscore in it, and I think that the `_ci` part might just be unnecessary. What do you think about shortening it? If not shortening, what about changing it to `diffux-ci` or `diffuxci`?

trotzig:
When I chose the name for the command-line tool, I thought that
diffux-core [1] had already occupied that name. But it turns out I was
wrong, diffux-core uses `diffux-snapshot` and `diffux-compare`.

Fixes #43

[1]: https://github.com/diffux/diffux-core


lencioni:
Looks good to me. After we release the new version, we'll need to update brigade and probably want to update the blog post as well.

trotzig:
Yeah, I figured we would have a bunch of things in the release to take care
of in the brigade repo. Good call about the blog post too!

PR #50, opened 2015-10-25
Pass list of examples from JS to Ruby

trotzig:
Instead of iterating blindly until we run out of examples, we now first
get a list of all examples and then use that list to iterate. This will
help an upcoming change to parallellize execution.

Since I changed the array of defined examples into a hash/object, I had
to move the duplicate description introduced in e45442a into javascript
land.


lencioni:
Good work! This will also make it easy for us to have multiple `fdefine`s, which is something I've been wanting lately. I'll add that now.

PR #51, opened 2015-10-25
Make `fdefine` work for multiple examples simultaneously

lencioni:
I sometimes want to focus on multiple examples at the same time. Now that we compile a list of descriptions, expanding `fdefine` to allow for multiple examples is fairly trivial. This also matches RSpec and Jasmine behavior, which is more likely what people will expect anyway.


trotzig:
Looks good!

PR #52, opened 2015-10-25
Improve output while rendering examples

lencioni:
I found that when an error happened, it wasn't exactly clear which
example it happened in from the output. To make this better, I decided
to break up the output from all happening in one place to happening in
multiple steps. This allows us to render the description of the example
on its own line, and then the viewport information separately, which
should help us pinpoint errors a little faster.

While I was at it, I added some information to the viewport output and
added some dots to be rendered incrementally as we progress through the
rendering. This will give people a better sense that something is
happening while running diffux.

### DIFF
--- a/lib/diffux_ci_runner.rb
+++ b/lib/diffux_ci_runner.rb
@@ -54,11 +54,21 @@ begin
 
   all_examples.each do |example|
     description = example['description']
+    puts description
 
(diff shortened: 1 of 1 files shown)


trotzig:
Looks good. I think this will help make running the command feel smoother.

PR #56, opened 2015-10-25
Simplify clearing out document between examples

lencioni:
We were previously just removing any visible elements from the body,
which requires a little overhead to determine if the element is visible.
I believe we were doing this to avoid removing our script tags which
were in the body. I think that if we move the script tags to the head,
we can do this more simply.

In my unscientific test, this didn't seem to have much of an impact on
performance, but it is simpler so I think it is still worth doing.

### DIFF
--- a/lib/public/diffux_ci-runner.js
+++ b/lib/public/diffux_ci-runner.js
@@ -49,23 +49,6 @@ window.diffux = {
     }.bind(this));
   },
 
-  isElementVisible: function(element) {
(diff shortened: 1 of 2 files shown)


trotzig:
If I'm not mistaken, I did this because the <Icon> component in the Brigade repo was depending on a hidden element being in the DOM (a blob of svg). Since we've moved to a different approach, removing everything was a good call. Though I have a feeling it might come back and bite at some point. But better to re-apply the old behavior if that happens rather than keeping it around.

PR #57, opened 2015-10-25
Simplify fdefine filtering

lencioni:
I realized that we were filtering the descriptions down to be the same
array as `this.fdefine`, so we can just totally skip that step and just
use `this.fdefine` instead.


trotzig:
:+1:

PR #58, opened 2015-10-25
Make sure renderExample() errors gracefully

lencioni:
We call this function asynchronously from Ruby. In order for it to know
what is happening, we need to make sure that doneFunc() is called. Since
we already have a catch block wired up, we can just move our throw into
there and we should be good to go.

I added a comment to help explain this bit of weirdness for the future.
I also decided to throw a new Error object to give us more context here,
like a stack trace, if we end up wanting that.

### DIFF
--- a/lib/public/diffux_ci-runner.js
+++ b/lib/public/diffux_ci-runner.js
@@ -96,17 +96,21 @@ window.diffux = {
   },
 
   /**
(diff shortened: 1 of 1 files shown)


trotzig:
Whoops, I merged the PRs in the wrong order. I'll merge locally and re-push.

PR #59, opened 2015-10-26
Avoid cropping unless we actually need it

lencioni:
Under the hood, this will cause ChunkyPNG to allocate a new array of
pixels for the cropped image, and then slice out the pixels needed for
the cropped image from the uncropped image and put them into the new
array. If we don't actually need to crop, we can avoid this work and
memory allocation.

In the Brigade diffux test suite this seemed to happen a few times, but
not enough to make a noticeable difference in performance. Still, since
it is such an easy thing to do, we might as well do it.


trotzig:
:+1:

PR #70, opened 2015-12-17
Reorganize project

lencioni:
To more closely follow conventions for how a gem's files are organized,
I am moving some things under lib/diffux_ci and the DiffuxCI module.


trotzig:
:+1:

PR #80, opened 2016-07-08
Improve error stack trace

serena-w:
This change allows the user to better see where the error came from. The
stack trace helps the user to more easily debug the error.


trotzig:
Looks good to me. Thanks @serena-w for contributing to diffux_ci!

PR #85, opened 2016-07-09
Prevent newlines in directory names

### LINKED ISSUE #76: Directory names contain newlines
Because of how [`Base64.encode64`](http://ruby-doc.org/stdlib-2.1.3/libdoc/base64/rdoc/Base64.html#method-i-encode64) works. Might be better to use `strict_encode64`.

(This tripped up CircleCI when trying to export the files as artefacts.)

lencioni:
It turns out that Base64.encode64 adds newlines after a certain length.
This ends up causing problems when used in filenames. Instead, we can
use Base64.strict_encode64, which does the same thing but without
newlines.

Fixes #76.

### DIFF
--- a/lib/diffux_ci/utils.rb
+++ b/lib/diffux_ci/utils.rb
@@ -35,7 +35,7 @@ module DiffuxCI
     end
 
     def self.normalize_description(description)
-      Base64.encode64(description).strip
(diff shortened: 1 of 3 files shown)


trotzig:
Thanks!

janpaul123:
Thanks indeed!

PR #86, opened 2016-07-12
Allow additional directories to be specified in the config

### LINKED ISSUE #82: Provide a way to include custom fonts
If a component's styles references a custom font using `@font-face` or whatnot, there should be a way for diffux to be configured to be able to serve up the required font files. I'm thinking something similar to the `stylesheets` configuration in the YAML file.

If there is a way to do this already, it should be documented.

cc @serena-w

serena-w:
I added a way for diffux to be configured to look for resources in other
directories. This allows the usage of custom fonts and resolves the issue
https://github.com/diffux/diffux_ci/issues/82. I also added a test to
check that diffux can find files in other directories specified in its
config.

fixes #82

### DIFF
--- a/lib/diffux_ci/server.rb
+++ b/lib/diffux_ci/server.rb
@@ -38,6 +38,21 @@ module DiffuxCI
       end
     end
 
+    get '/*' do
+      config = DiffuxCI::Utils.config
+      file = params[:splat].first
+      if File.exist?(file)
+        send_file file
+      else
+        config['public_directories'].each do |pub_dir|
+          filepath = File.join(Dir.pwd, pub_dir, file)
+          if File.exist?(filepath)
+            send_file filepath
+          end
+        end
+      end
+    end
+
     post '/reject' do
       DiffuxCI::Action.new(params[:description], params[:viewport]).reject
       redirect back
--- a/spec/diffux_ci_spec.rb
+++ b/spec/diffux_ci_spec.rb
@@ -523,4 +523,74 @@ describe 'diffux_ci' do
         .to eq(false)
     end
   end
+
+  describe 'when additional files are served from public directories' do
+    before do
+      tmp_pub_dir = File.join(@tmp_dir, 'public')
+      Dir.mkdir(tmp_pub_dir)
+
+      File.open(File.join(tmp_pub_dir, 'picture.gif'), 'wb') do |f|
+        tiny_gif = 'R0lGODlhAQABAIABAP///wAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw=='
+        f.write(Base64.decode64(tiny_gif))
+      end
+    end
+
+    let(:config) do
+      {
+        'source_files' => ['examples.js'],
+        'public_directories' => ['public']
+      }
+    end
+
+    let(:examples_js) { <<-EOS }
+      diffux.define('img', function() {
+        return new Promise(function(resolve, reject) {
+          var image = new Image();
+          image.onload = function() {
+            // Continue to process the image once it is found without any errors
+            resolve(image);
+          };
+          image.onerror = function() {
+            // Throws an error if the image is not found.
+            // The error message will then show up in std_err, so for our test,
+            // we can check that the error message should not show up.
+            reject(new Error('image not found'));
+          };
+          image.src = 'picture.gif';
+          document.body.appendChild(image);
+        });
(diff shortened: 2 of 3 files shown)


lib/diffux_ci/server.rb
```diff
@@ -38,6 +38,27 @@ def h(text)
       end
     end
 
+    get '/*' do
+      config = DiffuxCI::Utils.config
+      current_dir = '/Users/serena_wang/Documents/diffux_ci/lib/diffux_ci'
```
lencioni: This needs to be changed

lib/diffux_ci/server.rb
```diff
@@ -38,6 +38,27 @@ def h(text)
...
+      File.open(File.join(current_dir,'output.txt'), 'w') do |f|
+        f.write(Dir.pwd)
```
lencioni: I think you left this in from debugging

lib/diffux_ci/server.rb
```diff
@@ -38,6 +38,27 @@ def h(text)
...
+      end
+      file = params[:splat].first
+      if File.exist?(file)
+        send_file file
+      elsif File.exist?(File.join(Dir.pwd, file))
+        send_file File.join(Dir.pwd, file)
+      else
```
lencioni: Do you need both the `if` and `elsif` parts of this route? It seems like you might not.

spec/diffux_ci_spec.rb
```diff
@@ -523,4 +523,65 @@ def snapshot_file_exists?(description, size, file_name)
...
     end
   end
+
+  describe 'when additional files are served from public directories' do
+    before do
+      @tmp_pub_dir = File.join(@tmp_dir, 'public')
+      Dir.mkdir(@tmp_pub_dir)
+
+      File.open(File.join(@tmp_pub_dir, 'picture.gif'), 'wb') do |f|
+        f.write(Base64.decode64('R0lGODlhAQABAIABAP///wAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw=='))
```
lencioni: It would be nice to move this base64 encoded string into a well named variable to make this code more readable.

spec/diffux_ci_spec.rb
```diff
@@ -523,4 +523,65 @@ def snapshot_file_exists?(description, size, file_name)
...
+    let(:config) do
+      {
+        'source_files' => ['examples.js'],
+        'public_directories' => ['public']
+      }
+    end
+
+    let(:examples_js) { <<-EOS }
+      diffux.define('img', function() {
+        return new Promise( function(resolve, reject) {
```
lencioni: nit: no space between `(` and `function`

spec/diffux_ci_spec.rb
```diff
@@ -523,4 +523,65 @@ def snapshot_file_exists?(description, size, file_name)
...
+          var image = new Image();
+          image.onload = function() {
+            resolve(image);
+          };
+          image.onerror = function() {
+            reject(new Error('image not found'));
```
lencioni: Our approach here is a little unusual, so it is probably worth documenting why we are doing this weird stuff with a comment.

lencioni:
It might be good to document some of the approaches you considered or tried in your commit message and why you landed on the catch-all route.

lencioni:
It also looks like your test did not pass in CI, FYI.

spec/diffux_ci_spec.rb
```diff
@@ -523,4 +523,74 @@ def snapshot_file_exists?(description, size, file_name)
         .to eq(false)
     end
   end
+
+  describe 'when additional files are served from public directories' do
+    before do
+      @tmp_pub_dir = File.join(@tmp_dir, 'public')
```
lencioni: Since this variable isn't used outside of this block, you don't need the `@`. If you remove it, it will make your code a little clearer.

lencioni:
one minor comment, otherwise this looks good. Thanks for working on this!

trotzig:
Thanks @serena-w! This is a great addition!

trotzig:
I just have to say a big THANK YOU for this change. We recently switched font-loading strategy at brigade.com (to [the "Critical FOFT with data URI" technique](https://www.zachleat.com/web/comprehensive-webfonts/#critical-foft-data-uri)), and I was able to make our happo run happy by simply adding a line to our config. Thanks!

Issue #81, opened 2016-07-08
Rewrite in JavaScript


lencioni:
I've been floating this idea around in my mind for a while, and I think it is time to open it up for discussion. Ruby has served this project well, but I think if we want to take it to the next level, we should consider rewriting in JavaScript at some point.
## Advantages
- Lower barrier to contributing from front-end developers
- Lower barrier to using by front-end developers
- Reduces the number of technologies required by build systems
- A good chunk of this is already written in JavaScript, so it would reduce the awkwardness of the interaction layer that we currently have between Ruby and JS
- The dependency on a diffux package could be made explicit in the tests, instead of relying on globals like `diffux.describe` (e.g. `import { describe } from 'diffux';`)
- Would be easy to set up a build step so we can take advantage of Babel in this project
## Disadvantages
- We already have a good and fast way to generate PNGs in Ruby/C, and JS packages for this are unknown to me (there is likely something good)
- Similarly, we already have a good way to perform diffs on the screenshots in Ruby, and JS packages for this are unknown to me
## Other considerations
- We have talked about retiring the diffux_core gem and moving the relevant code into this project #79, that might be a natural time to convert to JS.
- We've also talked about generating diff images that make more sense to newcomers, and possibly improving the UI for reviewing diffs #61. It might be a natural time to make these changes when we convert to JS.
- We recently did a similar rewrite of import-js from Ruby to JavaScript, so we have some experience with it. It also seems to have been a good choice for that project.

trotzig:
I'm definitely pro this, although I don't have the same bandwidth converting diffux as I had for import-js.

lencioni:
That's fair. Me too, unless I am able to do this during the day. I think this conversion may be easier anyway--we certainly have a lot fewer tests to convert.

lencioni:
I'm doing a bit of research on this. I think the biggest open questions for me is what will we replace our dependencies with? Some of them seem like they will have pretty straightforward or obvious replacements (e.g. s3, selenium-webdriver, sinatra (I was thinking express)). So I've been looking into what we might replace chunky PNG and diff-lcs with.

I discovered [png-diff](https://github.com/chenglou/png-diff) which appears to do some of the same things we'll need to do here. This lead me to [pngjs](https://github.com/lukeapage/pngjs) for reading and writing PNG files. Seems promising.

[adiff](https://github.com/dominictarr/adiff) might be worth looking at to calculate the diffs, but I'm not sure if it is able to operate on arrays or 2D arrays.

[...]

Issue #79, opened 2016-07-07
Move to Galooshi


lencioni:
I think we should:

- [x] mark [diffux](https://github.com/diffux/diffux) as deprecated
- [x] retire the [diffux core gem](https://github.com/diffux/diffux-core)
- [x] move the relevant diffux core code into this project
- [x] rename this project to "Happo"
- [x] move this project to the Galooshi organization

lencioni:
I asked around for new names, here's a few of the better ones we came up with:

- diffooey (diffUI)
- pixelwatch
- odo (the shapeshifter, unfortunately taken on npm)
- gaius (as in [Gaius Baltar](https://en.wikipedia.org/wiki/Gaius_Baltar))

I am currently leaning toward gaius. What do you think?

trotzig:
Maybe we should just call it Diffui? It can be pronounced diffooey. Or maybe that's what you suggested? I'm neutral on the Gaius idea, I'm not a mega fan of Battlestar Galactica. 

Alternatively (and half-seriously, half-jokingly), my nephew just came up with "Happo Carappo". He gave me a drawing on my birthday with a glued-on airplane and the text "Happo Carappo, here comes the Söderkvist people". He's three years old.

lencioni:
How about just Happo? I like that!

trotzig:
Happo sounds like a startup, so yeah!

lencioni:
It fits in well with "Galooshi" too. :)

lencioni:
If you still have the drawing, I'd love to see a picture of it here.

trotzig:
Unfortunately not :(

lencioni:
I added a deprecation notice to diffux core in https://github.com/diffux/diffux-core/commit/32febef8a6ea850074da4e9848d34681a01bc1b6

PR #87, opened 2016-07-13
Move code over from diffux-core

lencioni:
I am working on moving this project to the Galooshi organization (#79) and setting us up to rewrite in JavaScript (#81). As part of this work, I want to simplify things by moving the code we use from the diffux-core gem into this project. Since we have deprecated Diffux and diffux-core, this seems like a good move to me and it will make this project more self-contained and easier to work with.

I approached this by moving the files over, and then simplifying them to make the code easier to track.


trotzig:
LGTM

PR #89, opened 2016-07-14
Remove approve/reject feature

### LINKED ISSUE #71: Approving/rejecting a diff from the Review page appears to be broken
It's actually not broken, but when you reject/approve we re-render the page using information in the `result_summary.yaml` which is never updated.

lencioni:
We brought this idea over from the original diffux project. However, as
we've used this tool, we have discovered that this feature is rarely
used and generally not that useful in this context. Most of the time
this will be run in CI where having approve/reject is just a little
confusing. To keep this project focused and reduce the amount of code we
need to maintain, I have decided to remove this feature.

We can likely simplify how we store files and tweak some of the
interface copy, but I'm going to leave that for another commit.

Fixes #71.


trotzig:
Thanks for doing this. It looks good, though I think we're leaving users in a bad situation if they want to run it locally, because we will always compare to the baseline (which can't be changed when there are no reject/accept buttons).

trotzig:
I think we should:
- Keep two snapshots - `current` and `previous`
- Move `current` to `previous` when we snapshot a new `current` 

What do you think?

lencioni:
Ah, that's right. I'm working on renaming things right now. I'll next work on moving images around.

trotzig:
Cool. Feel free to merge this PR if it helps you move forward.

lencioni:
One bummer about this is that we currently save a little work when there is no diff because we don't have to write out the current snapshot to disk or do anything with the baseline. By changing to this, we will need to write something to disk every time I think, which will definitely affect performance.

One thought I had to get around this is to:

1. Delete previous.png if it exists.
2. Compare the current snapshot in memory against current.png if it exists.
3. If there is no diff, return, leaving the old current.png in place.
4. If there is a diff, move current.png to previous.png and write out diff.png to disk and the current snapshot to current.png.

This would allow us to keep the current performance characteristics of this operation. What do you think?

trotzig:
sounds good.

lencioni:
Alright! I've pushed that change for review. I likely broke some specs and there are maybe actual bugs in the code, but I am out of time right now. Feel free to review or fix up any mistakes.

lencioni:
Spec has been updated. I believe this is ready.

happo_example.sh
```diff
@@ -16,7 +16,7 @@ run-happo() {
   xvfb-run happo
 }
 
-# Check out the previous version and generate baseline snapshots
+# Check out the previous version and generate previous snapshots
```
trotzig: This sentence is a little strange. Maybe you can just leave out previous from the second part:

"Check out the previous version and generate snapshots"

trotzig:
LGTM

happo_example.sh
lencioni: caf25dd

lencioni:
I believe this is a breaking change, so the next version should be 2.0.0.

Issue #91, opened 2016-07-15
Allow secret S3 credentials to be specified via environment variables


lencioni:
Right now, this project requires that the S3 credentials are in your YAML file.

```yaml
s3_access_key_id: <your acccess key id>
s3_secret_access_key: <your secret acccess key>
s3_bucket_name: <a globally unique bucket name>
```

In the example project, [we use Ruby to dynamically read these values in, which is a bit weird](https://github.com/Galooshi/happo/blob/ce49454700aecfede3950298bf85d25ac7b34658/example-project/.happo.yaml#L14-L15).

```yaml
s3_access_key_id: <%= ENV['S3_ACCESS_KEY_ID'] %>
s3_secret_access_key: <%= ENV['S3_SECRET_ACCESS_KEY'] %>
s3_bucket_name: <%= ENV['S3_BUCKET_NAME'] %>
```

Perhaps we should remove these from the configuration file and always require them to be specified via an environment variable?

trotzig:
Sounds good to me.

lencioni:
This was implemented by #92

PR #92, opened 2016-07-20
Make S3 credentials be specified by environment variables

serena-w:
The S3 credentials used for uploading diff images are now required to be
specified by environment variables, instead of in the `.happo.yaml`
file. Users must set `S3_ACCESS_KEY_ID`, `S3_SECRET_ACCESS_KEY`, and
`S3_BUCKET_NAME` as environment variables to use `happo upload_diffs`.
Users can also set `S3_BUCKET_PATH` as an environment variable if they
wish to specify a directory path for where they want diff images
uploaded within the S3 bucket. This also addresses https://github.com/Galooshi/happo/issues/91.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,9 +149,6 @@ viewports:
   small:
     width: 320
     height: 444
-s3_access_key_id: <your acccess key id>
-s3_secret_access_key: <your secret acccess key>
-s3_bucket_name: <a globally unique bucket name>
 ```
 
 ## Command line tools
@@ -177,8 +174,25 @@ and you can do use your developer tools to debug.
 ### `happo upload_diffs`
 
 Uploads all current diff images to an Amazon S3 account and reports back URLs
-to access those diff images. Requires the `s3_access_key_id`,
-`s3_secret_access_key`, and `s3_bucket_name` configuration options.
+to access those diff images. Requires that `S3_ACCESS_KEY_ID`,
+`S3_SECRET_ACCESS_KEY`, and `S3_BUCKET_NAME` are specified as environment
+variables. `S3_ACCESS_KEY_ID` and `S3_SECRET_ACCESS_KEY` will be the
+credentials Happo uses to access the bucket named `S3_BUCKET_NAME`.
+`S3_BUCKET_PATH` can also be set as an environment variable to specify a
+directory path for where you want diff images uploaded within the S3 bucket.
+You can set these in the session by using `export`:
+```sh
+export S3_ACCESS_KEY_ID=<YOUR_ACCESS_KEY_VALUE>
+export S3_SECRET_ACCESS_KEY=<YOUR_SECRET_ACCESS_KEY_VALUE>
+export S3_BUCKET_NAME=<YOUR_BUCKET_NAME>
+export S3_BUCKET_PATH=<YOUR_BUCKET_PATH>
+
+happo upload_diffs
+```
+or by adding them in the beginning of the command:
+```sh
+S3_ACCESS_KEY_ID=<...> S3_SECRET_ACCESS_KEY=<...> ... happo upload_diffs
+```
 
 ### `happo clean`
 
--- a/lib/happo/uploader.rb
+++ b/lib/happo/uploader.rb
@@ -4,9 +4,10 @@ require 'securerandom'
 module Happo
   class Uploader
     def initialize
-      @s3_access_key_id = Happo::Utils.config['s3_access_key_id']
-      @s3_secret_access_key = Happo::Utils.config['s3_secret_access_key']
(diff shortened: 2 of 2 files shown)


lib/happo/uploader.rb
```diff
@@ -18,6 +19,9 @@ def upload_diffs
 
       bucket = find_or_build_bucket
       dir = SecureRandom.uuid
+      if !(@s3_bucket_path.nil? || @s3_bucket_path.empty?)
+        dir = File.join(@s3_bucket_path, SecureRandom.uuid)
+      end
```
lencioni: We like for our commits to be small and atomic. It would have been nice to keep this change that adds the S3_BUCKET_PATH in a separate commit. Don't worry about splitting it up now, but please keep it in mind for next time.

lib/happo/uploader.rb
```diff
@@ -18,6 +19,9 @@ def upload_diffs
 
       bucket = find_or_build_bucket
       dir = SecureRandom.uuid
+      if !(@s3_bucket_path.nil? || @s3_bucket_path.empty?)
+        dir = File.join(@s3_bucket_path, SecureRandom.uuid)
```
lencioni: Instead of generating a second uuid, we might as well reuse the one we already generated on line 21. Alternatively, this code could be refactored so that we only set `dir` once (e.g. by using an `else`)

lencioni: Using an `else` would also allow you to make the condition positive, which is easier to grok.

```rb
dir = if @s3_bucket_path.nil? || @s3_bucket_path.empty?
        SecureRandom.uuid
      else
        File.join(@s3_bucket_path, SecureRandom.uuid)
      end
```

README.md
```diff
@@ -177,8 +174,9 @@ and you can do use your developer tools to debug.
 ### `happo upload_diffs`
 
 Uploads all current diff images to an Amazon S3 account and reports back URLs
-to access those diff images. Requires the `s3_access_key_id`,
-`s3_secret_access_key`, and `s3_bucket_name` configuration options.
+to access those diff images. Requires that `S3_ACCESS_KEY_ID`,
+`S3_SECRET_ACCESS_KEY`, `S3_BUCKET_NAME`, and `S3_BUCKET_PATH` are specified
+as environment variables.
```
lencioni: This isn't quite accurate since S3_BUCKET_PATH is not required.

It would be good to add some information about what each of these environment variables does and also show some code that demonstrates how to set them.

lib/happo/uploader.rb
serena-w: so something like
```
if @s3_bucket_path.nil? || @s3_bucket_path.empty?
  dir = SecureRandom.uuid
else
  dir = File.join(@s3_bucket_path, SecureRandom.uuid)
end
```

lencioni: Sure, or you could use the code I posted in https://github.com/Galooshi/happo/pull/92#discussion_r71615131 :D

README.md
serena-w: Addressed in the latest version, PTAL

README.md
```diff
@@ -177,8 +174,19 @@ and you can do use your developer tools to debug.
...
 Uploads all current diff images to an Amazon S3 account and reports back URLs
-to access those diff images. Requires the `s3_access_key_id`,
-`s3_secret_access_key`, and `s3_bucket_name` configuration options.
+to access those diff images. Requires that `S3_ACCESS_KEY_ID`,
+`S3_SECRET_ACCESS_KEY`, and `S3_BUCKET_NAME` are specified as environment
+variables. `S3_ACCESS_KEY_ID` and `S3_SECRET_ACCESS_KEY` will be the
+credentials Happo uses to access the bucket named `S3_BUCKET_NAME`.
+`S3_BUCKET_PATH` can also be set as an environment variable to specify a
+directory path for where you want diff images uploaded within the S3 bucket.
+You can set these in your shell or ~/.bashrc using `export`.
```
lencioni: I think the detail about the `.bashrc` file can be left out, since this will usually be run in CI where you likely won't have access to that file. Perhaps rephrase as:

```md
You can set these for the session by using `export`:

export S3_ACCESS_KEY_ID=<YOUR_ACCESS_KEY_VALUE>
export S3_SECRET_ACCESS_KEY=<YOUR_SECRET_ACCESS_KEY_VALUE>
export S3_BUCKET_NAME=<YOUR_BUCKET_NAME>
export S3_BUCKET_PATH=<YOUR_BUCKET_PATH>

happo upload_diffs

or by adding them to the beginning of the command:

S3_ACCESS_KEY_ID=<...> S3_SECRET_ACCESS_KEY=<...> ... happo upload_diffs
```

PR #93, opened 2016-07-21
Get single S3 bucket instead of all buckets

serena-w:
Asking for all buckets may lead to access denied errors if the IAM role or user
associated with the provided access key pair does not have permissions for all
buckets. Furthermore, we only need to retrieve the bucket specified by the
environment variable `S3_BUCKET_NAME` to put diff images into, and we should
definitely have access to `S3_BUCKET_NAME` with the provided S3 credentials. Thus, we should use `service.bucket(@s3_bucket_name)` instead of `service.buckets.find(@s3_bucket_name)`.

For some reason, the `service.bucket("bucket_name")` method was not specified
in the README for the `s3` library, and the only example in the documentation
for finding a single bucket was `service.buckets.find("bucket_name")`. I have
submitted a pull request to update the documentation there as well: https://github.com/qoobaa/s3/pull/116. Also, since
`service.bucket(...)` doesn't check whether the bucket with the given name
exists, we should use the provided `exists?` method from the library to check
whether the bucket actually exists.

### DIFF
--- a/lib/happo/uploader.rb
+++ b/lib/happo/uploader.rb
@@ -62,9 +62,9 @@ module Happo
     def find_or_build_bucket
       service = S3::Service.new(access_key_id: @s3_access_key_id,
                                 secret_access_key: @s3_secret_access_key)
-      bucket = service.buckets.find(@s3_bucket_name)
+      bucket = service.bucket(@s3_bucket_name)
 
-      if bucket.nil?
+      if !bucket.exists?
         bucket = service.buckets.build(@s3_bucket_name)
         bucket.save(location: :us)
       end


lencioni:
Thanks @serena-w! Can you include a link to your PR on the s3 project here for reference?

serena-w:
@lencioni I've updated the PR description with a link to the PR for updating the S3 project documentation now.

lib/happo/uploader.rb
```diff
@@ -62,9 +62,9 @@ def upload_diffs
     def find_or_build_bucket
       service = S3::Service.new(access_key_id: @s3_access_key_id,
                                 secret_access_key: @s3_secret_access_key)
-      bucket = service.buckets.find(@s3_bucket_name)
+      bucket = service.bucket(@s3_bucket_name)
 
-      if bucket.nil?
+      if !bucket.exists?
         bucket = service.buckets.build(@s3_bucket_name)
```
trotzig: Is there a similar way to avoid fetching all buckets here too? Something like `service.build_bucket`. Or perhaps this call won't cause the error that we saw above?

trotzig:
This is great, thanks @serena-w!

lib/happo/uploader.rb
lencioni: Great question. I looked at the S3 code and it looks like `service.buckets` will still end up causing the error here. However, `service.buckets.build` looks like it just instantiates a new `Bucket`, which `service.bucket` also does, so we can potentially just remove this line now. The only thing I'm not sure about is the `proxy_owner` argument.

@serena-w can you try that out and submit a PR if it works?

https://github.com/qoobaa/s3/blob/ec570632d25a3571cd4585de08e48e72b838f048/lib/s3/buckets_extension.rb#L5

serena-w: The `proxy_owner` argument in `service.buckets.build` seems to be just `service`, which is passed into BucketExtensions as `self` in `service.buckets` (https://github.com/qoobaa/s3/blob/ec570632d25a3571cd4585de08e48e72b838f048/lib/s3/service.rb#L48). `service.bucket` also creates a bucket with `self` for the same argument (https://github.com/qoobaa/s3/blob/ec570632d25a3571cd4585de08e48e72b838f048/lib/s3/service.rb#L55). Thus, the bucket `service.buckets.build` creates is actually identical to a bucket created by `service.bucket`. I will submit a PR removing this line in `uploader.rb`.

Issue #95, opened 2016-07-22
Doesn't send 404s for files that are not found


lencioni:
I believe that the change from #86 made it so that files that are not found will not result in 404 errors. I'm not a Sinatra expert, but I think we need to add something like

```rb
halt 404, "Not found"
```

if we can't find the file.

cc: @serena-w

trotzig:
I'm guessing this doesn't really affect the general `happo` run in any significant way, right?

lencioni:
That's right. The main impact it has is if you are running happo in debug mode and are using the new public directory setting and don't have things set up quite right so the files aren't there, instead of 404s you'll get 200s.

lencioni:
This was fixed by #96.

PR #96, opened 2016-07-27
Show 404 error for file not found

serena-w:
This commit deals with this issue
https://github.com/Galooshi/happo/issues/95. Now, if you run `happo
debug` and try to go to a file that cannot be found, a 404 error is
raised. Previously, you would have gotten a 202 error if a file could
not be found. I missed this bug in
https://github.com/Galooshi/happo/pull/86.

### DIFF
--- a/lib/happo/server.rb
+++ b/lib/happo/server.rb
@@ -43,14 +43,18 @@ module Happo
       file = params[:splat].first
       if File.exist?(file)
         send_file file
+        return
       else
         config['public_directories'].each do |pub_dir|
           filepath = File.join(Dir.pwd, pub_dir, file)
           if File.exist?(filepath)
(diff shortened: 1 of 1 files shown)


lib/happo/server.rb
```diff
@@ -43,14 +43,18 @@ def h(text)
       file = params[:splat].first
       if File.exist?(file)
         send_file file
+        return
       else
```
trotzig: This is a minor issue, but since we now return in the if-statement above, we don't need the `else`.

trotzig:
Looks good to me. Thanks!

lib/happo/server.rb
lencioni: I'll follow up with some clean up. :)

PR #97, opened 2016-07-27
Clean up catchall route

lencioni:
Now that we are using early returns in this codepath, we can flatten the
logic a bit to improve readability of the code. While I was at it, I
added some comments explaining what is going on here and why it is
written the way it is, to help future explorers.

### DIFF
--- a/lib/happo/server.rb
+++ b/lib/happo/server.rb
@@ -39,20 +39,20 @@ module Happo
     end
 
     get '/*' do
-      config = Happo::Utils.config
+      # If this file is part of Happo itself (e.g. some bit of JavaScript), we
+      # want to give that precedence so that Happo can work properly.
       file = params[:splat].first
-      if File.exist?(file)
-        send_file file
-        return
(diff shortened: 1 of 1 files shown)


lib/happo/server.rb
```diff
@@ -39,20 +39,20 @@ def h(text)
...
-          filepath = File.join(Dir.pwd, pub_dir, file)
-          if File.exist?(filepath)
-            send_file filepath
-            return
-          end
-        end
+      return send_file(file) if File.exist?(file)
+
+      # The requested file is not part of Happo, so let's look in the configured
+      # public directories. This is useful for serving up custom font files that
```
trotzig: I was expecting a `e.g.` here: "This is useful for serving up e.g. custom font files that ..."

trotzig:
lgtm!

lib/happo/server.rb
lencioni: done!

PR #98, opened 2016-07-28
Use relative paths instead of absolute paths for images

serena-w:
With this change, the image elements in the generated `index.html` for
`happo upload_diffs` will use relative paths instead of absolute paths
for their source urls.

In my case, I'm using a proxy to allow developers without access to the
S3 bucket containing the image and HTML files Happo creates to also see
the visual diffing images. I would like to use the generated
`index.html` to present the visual diffing images, but the browser would
not have the access to get to the absolute paths leading to the S3 bucket for
the images. Using relative paths allows the images to be retrieved using
the proxy as well.

### DIFF
--- a/lib/happo/uploader.rb
+++ b/lib/happo/uploader.rb
@@ -25,26 +25,26 @@ module Happo
             end
 
       diff_images = result_summary[:diff_examples].map do |diff|
-        image = bucket.objects.build(
-          "#{dir}/#{diff[:description]}_#{diff[:viewport]}.png")
+        img_url = "#{dir}/#{diff[:description]}_#{diff[:viewport]}.png"
+        image = bucket.objects.build(img_url)
         image.content = open(Happo::Utils.path_to(diff[:description],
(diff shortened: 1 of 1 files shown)


lencioni:
Awesome! This is great. Thanks!

lib/happo/uploader.rb
```diff
@@ -25,26 +25,26 @@ def upload_diffs
...
         image.save
-        diff[:url] = image.url
+        diff[:url] = img_url
         diff
       end
 
       new_images = result_summary[:new_examples].map do |example|
-        image = bucket.objects.build(
-          "#{dir}/#{example[:description]}_#{example[:viewport]}.png")
+        img_url = "#{dir}/#{example[:description]}_#{example[:viewport]}.png"
```
lencioni: I should have noticed this before, but these URLs won't work correctly, since they begin with the directory name and not `/`. I think we can shorten these to be just the filename portion for the page, so they are relative to the page.

PR #100, opened 2016-07-29
Fixed uploader bug for new images

serena-w:
Previously, `happo upload_diffs` would throw an error every time
the `previous.png` image for a test could not be found. Thus, if a new
test was defined or if the component didn't change, `happo upload_diffs`
would throw an error. For new images with no `previous.png` or
`diff.png`, the uploader should upload `current.png`.


lencioni:
Thanks! This got messed up when we removed the accept/reject buttons.

PR #101, opened 2016-08-05
Escape unsafe characters in example descriptions

serena-w:
An example description may contain unsafe characters like quotation
marks or other punctuation marks that would break the URL used for the
image source in the generated `index.html`. We need to escape unsafe
characters in example descriptions so that the image source URL always
works correctly in the page displaying diff images and new examples.


trotzig:
Thanks!

PR #109, opened 2016-08-15
Include margin of topmost element in dimension calculations

### LINKED ISSUE #108: Include margin of topmost element in dimension calculations
We currently recurse through the DOM tree using `getBoundingClientRect()` to determine the shape of the box that we want to take a screenshot of. This is working well, except when the topmost element has margin on it, since `getBoundingClientRect()` does not include margin. This means that you might miss some important details about the visual appearance of your component if you aren't being super careful.

I think we should use `getComputedStyle()` to determine the margin of the topmost element and use that in our rectangle dimension calculations. I want to avoid doing this for every node in the tree, since it is not the fastest method. I think doing this on the topmost element will likely get us 99% of the benefit, so it seems like an easy win.

lencioni:
We currently recurse through the DOM tree using
`getBoundingClientRect()` to determine the shape of the box that we want
to take a screenshot of. This is working well, except when the topmost
element has margin on it, since `getBoundingClientRect()` does not
include margin. This means that you might miss some important details
about the visual appearance of your component if you aren't being super
careful.

I think we should use `getComputedStyle()` to determine the margin of
the topmost element and use that in our rectangle dimension
calculations. I want to avoid doing this for every node in the tree,
since it is not very fast. I think doing this on the topmost element
will likely get us 99% of the benefit, so it seems like an easy win.

Fixes #108.

cc: @serena-w

### DIFF
--- a/lib/happo/public/happo-runner.js
+++ b/lib/happo/public/happo-runner.js
@@ -182,6 +182,16 @@ window.happo = {
       top: rect.top,
     };
 
+    // getBoundingClientRect does not include margin, so we need to use
+    // getComputedStyle. Since this is slow and the margin of descendent
+    // elements is significantly less likely to matter, let's include the margin
+    // only from the topmost node.
+    var computedStyle = window.getComputedStyle(node);
+    box.bottom += parseInt(computedStyle.getPropertyValue('margin-bottom'), 10);
+    box.left -= parseInt(computedStyle.getPropertyValue('margin-left'), 10);
+    box.right += parseInt(computedStyle.getPropertyValue('margin-right'), 10);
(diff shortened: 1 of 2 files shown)


trotzig:
Looks good! Out of curiosity, does this mean that centered elements will include the full page width in their screenshots? (I'm assuming it does). 

```js
<div style="margin: 0 auto; width: 100px;">Centered</div>
```

lencioni:
I haven't tested it, but probably!

trotzig:
It's generally an anti-pattern to let the component itself handle its centering, so I'm not too worried about making screenshots too big.

PR #110, opened 2016-08-18
Fix snapshots that expand the top/left of the screen

trotzig:
In the brigade codebase, we have a component that renders 1px outside
the screen. The recent code we've added to improve the detection of the
box wasn't taking into account that elements might render off-screen.

Adding a simple check to see if either top or left is negative fixes the
issue. I didn't address this for elements that expand right or bottom of
the screen as I think the screenshotting process already handles those
(by including the content).

### DIFF
--- a/lib/happo/public/happo-runner.js
+++ b/lib/happo/public/happo-runner.js
@@ -162,6 +162,14 @@ window.happo = {
     box.right = Math.max(box.right, rect.right);
     box.top = Math.min(box.top, rect.top);
 
+    if (box.left < 0) {
+      box.left = 0; // Don't go outside window
+    }
+
+    if (box.top < 0) {
+      box.top = 0; // Don't go outside window
+    }
+
     for (var i = 0; i < node.children.length; i++) {
       this.getFullRectRecursive(node.children[i], box);
     }
--- a/lib/happo/public/happo-runner.js
+++ b/lib/happo/public/happo-runner.js
@@ -162,14 +162,6 @@ window.happo = {
     box.right = Math.max(box.right, rect.right);
     box.top = Math.min(box.top, rect.top);
 
-    if (box.left < 0) {
-      box.left = 0; // Don't go outside window
-    }
-
-    if (box.top < 0) {
-      box.top = 0; // Don't go outside window
-    }
-
     for (var i = 0; i < node.children.length; i++) {
       this.getFullRectRecursive(node.children[i], box);
     }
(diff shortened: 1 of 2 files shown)


trotzig:
ah, I should add a test for this

lib/happo/public/happo-runner.js
```diff
@@ -162,6 +162,14 @@ window.happo = {
     box.right = Math.max(box.right, rect.right);
     box.top = Math.min(box.top, rect.top);
 
+    if (box.left < 0) {
+      box.left = 0; // Don't go outside window
+    }
+
+    if (box.top < 0) {
+      box.top = 0; // Don't go outside window
+    }
```
lencioni: Am I understanding this correctly that without this, the Happo run errors out because it can't crop the screenshot to something off canvas?

lib/happo/public/happo-runner.js
```diff
@@ -162,6 +162,14 @@ window.happo = {
     box.right = Math.max(box.right, rect.right);
     box.top = Math.min(box.top, rect.top);
 
+    if (box.left < 0) {
+      box.left = 0; // Don't go outside window
```
lencioni: It would be better to do this at the very end, next to the width and height calculations, instead of doing this for every node.

Also, you can use `Math.max` if you like.

lib/happo/public/happo-runner.js
lencioni: Shouldn't that actually be handled by the code that does the cropping though?

trotzig: I figured it was better in the js file because then we don't have to recompute `box.width` etc in ruby. I actually started with a solution in rb but ended up here. I think you're right about not doing it for every node though. I'll move it outside of the recursion.

lib/happo/public/happo-runner.js
trotzig: You're right, I'll move these.

trotzig:
I pushed a new commit to move the box fixing outside the recursion. Thanks for reviewing @lencioni!

lib/happo/public/happo-runner.js
```diff
@@ -199,6 +199,11 @@ window.happo = {
       this.getFullRectRecursive(node.children[i], box);
     }
 
+    // Cut off things that render off the screen to the top or left, since those
+    // won't be in the screenshot file that we then crop from.
+    box.left = Math.max(box.left, 0);
+    box.top = Math.max(box.top, 0);
```
lencioni: And, we don't need to do this for the right or bottom because the screenshot will always include that, even if it goes off the page, right?

I wonder if this is actually a bug in webdriver...

lencioni:
LGTM

lib/happo/public/happo-runner.js
trotzig: Yeah. If this wasn't the behavior, I think a few of our current happo examples at brigade would fail. 

I'll add a comment. Probably after merging this PR to master though.

PR #112, opened 2016-08-23
Add favicon

trotzig:
Having a favicon will make it easier to find tabs open with happo diffs.
And it will make bookmarks look better I guess.

I just grabbed this image from the issue (#5), then ran it through
http://www.favicon-generator.org/ to generate an ico file. We should
update this image later on when we have a proper asset to work with.

I decided to use an inlined base64 image so that we don't have any
external dependencies for the html file we upload to S3.


trotzig:
I was going to do more than just the favicon, but for now I think this will do. I will open new PRs with other changes once I get to them.

lencioni:
Thanks!

PR #113, opened 2016-08-23
Improve debug

trotzig:


lib/happo/views/index.erb
```diff
@@ -15,7 +15,9 @@
       body {
         background-color: #fff;
         margin: 0;
-        pointer-events: none;
+        <% unless params[:debug] %>
+          pointer-events: none;
```
lencioni: It might be worth adding a comment here explaining why we do this in the first place.

lencioni:
LGTM

lib/happo/views/index.erb
trotzig: Good point.

Issue #103, opened 2016-08-07
Add way to preview uploads


lencioni:
It would be useful to have a way to simulate a diff upload and preview it locally without having to upload to S3. This would make it easier to improve that page.

PR #115, opened 2016-08-27
Merge diff views

trotzig:
To speed up future development on improving the diffs page, I've deleted the review template in favor of the diffs template. They've essentially been the same since we removed accept/reject.


trotzig:
I ended up pushing a minor enhancement to the diffs page as well.

lencioni:
This helps with #103 some.

lencioni:
I wonder if we should add a section at the end that lists (text-only) all of the examples without diffs?

lib/happo/public/happo-styles.css
```diff
@@ -1,4 +1,11 @@
 body {
   background-color: #f0f0f0;
   font-family: helvetica, arial;
+  background: repeating-linear-gradient(
+    -45deg,
+    #f0f0f0,
+    #f0f0f0 10px,
+    #f4f4f4 10px,
+    #f4f4f4 20px
+  );
```
lencioni: What do you think about a checkerboard pattern instead? That's traditionally what is used for this sort of thing.

```css
  background-color: #ebebeb;
  background-image: linear-gradient(45deg, #f4f4f4 25%, transparent 25%, transparent 75%, #f4f4f4 75%, #f4f4f4), 
  linear-gradient(45deg, #f4f4f4 25%, transparent 25%, transparent 75%, #f4f4f4 75%, #f4f4f4);
  background-size:20px 20px;
  background-position:0 0, 10px 10px
```

lencioni: Also, I think it might be nice to scope this pattern to a container that contains just the image

lib/happo/public/happo-styles.css
```diff
@@ -8,4 +8,27 @@ body {
     #f4f4f4 10px,
     #f4f4f4 20px
   );
+  margin: 0;
+}
+
+header {
```
lencioni: I know we aren't dealing with many pages, but I think these styles would be a lot better if we used BEM and avoided using element selectors entirely.

trotzig: Yeah, I thought about that. At this point it was just so simple that it didn't merit BEM. I saw you added it later though, that was a good call.

lib/happo/public/happo-styles.css
trotzig: I'll check out what you pushed soon. I tried a checkerboard first, but it looked like a grid and things weren't aligning well on top of it (headings mostly). .

PR #117, opened 2016-08-27
Add way to demo the review page, improve review page

### LINKED ISSUE #103: Add way to preview uploads

lencioni:
We want to vastly improve the diffs page, but that is slow going because
we don't have a super easy way to preview our changes. This commit fixes
this by adding a review-demo.rb script that starts up a server with some
dummy data in it so we can see what a diffs page might look like. This
should make it easier to improve this page.

Fixes #103

### DIFF
--- a/lib/happo/public/happo-styles.css
+++ b/lib/happo/public/happo-styles.css
@@ -24,7 +24,7 @@ body {
 .header {
   align-items: baseline;
   background-color: #ffffff;
-  border-bottom: 1px solid #cccccc;
+  box-shadow: 0 0 4px rgba(0, 0, 0, .3);
   display: flex;
   flex-wrap: wrap;
   justify-content: space-between;
--- a/lib/happo/server.rb
+++ b/lib/happo/server.rb
@@ -87,6 +87,11 @@ module Happo
           viewport: 'large',
           url: 'http://placehold.it/850x150',
         },
+        {
+          description: '<SomethingElseNew>',
+          viewport: 'small',
+          url: 'http://placehold.it/350x150',
+        },
       ]
 
       erb :diffs, locals: {
--- a/lib/happo/server.rb
+++ b/lib/happo/server.rb
@@ -52,6 +52,49 @@ module Happo
       }
     end
 
+    get '/review-demo' do
+      diff_images = [
+        {
+          description: 'First component',
+          viewport: 'small',
+          url: 'http://placehold.it/350x150',
+        },
+        {
+          description: 'First component',
+          viewport: 'medium',
+          url: 'http://placehold.it/550x150',
+        },
+        {
+          description: 'First component',
+          viewport: 'large',
+          url: 'http://placehold.it/850x150',
+        },
+      ]
+
+      new_images = [
+        {
+          description: 'New component',
+          viewport: 'small',
+          url: 'http://placehold.it/350x150',
+        },
+        {
+          description: 'New component',
+          viewport: 'medium',
+          url: 'http://placehold.it/550x150',
+        },
+        {
+          description: 'New component',
+          viewport: 'large',
+          url: 'http://placehold.it/850x150',
+        },
+      ]
+
+      erb :diffs, locals: {
+        diff_images: diff_images,
+        new_images: new_images
+      }
+    end
+
     get '/resource' do
       file = params[:file]
       if file.start_with? 'http'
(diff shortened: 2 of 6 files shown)


lencioni:
I also made some improvements to the diffs page in this PR.

lib/happo/server.rb
```diff
@@ -52,6 +52,49 @@ def h(text)
...
+          description: 'New component',
+          viewport: 'medium',
+          url: 'http://placehold.it/550x150',
+        },
+        {
+          description: 'New component',
+          viewport: 'large',
+          url: 'http://placehold.it/850x150',
+        },
+      ]
```
trotzig: This is good for documenting the structure of the objects we pass to the template as well.

lib/happo/utils.rb
```diff
@@ -32,7 +32,11 @@ def self.config
 
     def self.config_from_file
       config_file_name = ENV['HAPPO_CONFIG_FILE'] || '.happo.yaml'
-      YAML.load(ERB.new(File.read(config_file_name)).result)
+      if File.exist?(config_file_name)
+        YAML.load(ERB.new(File.read(config_file_name)).result)
+      else
+        {}
+      end
```
trotzig: Thanks for fixing, I ran into this issue yesterday.

review-demo.rb
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env ruby
+
+# Add the lib directory to Ruby's load path so our requires will work.
+$:.unshift(File.expand_path('./lib'))
+
+require 'happo'
+
+system 'open', Happo::Utils.construct_url('/review-demo')
+require 'happo/server'
```
trotzig: Did you consider making this part of the cli tool instead? `happo review-demo` or something.

lib/happo/views/diffs.erb
```diff
@@ -19,22 +19,26 @@
...
+          </h3>
 
-      <h2>New examples (<%= new_images.count %>)</h2>
-      <% new_images.each do |image| %>
-        <h3>
-          <%= Rack::Utils.escape_html(image[:description]) %> @ <%= image[:viewport] %>
-        </h3>
-        <p><img src="<%= image[:url] %>"></p>
+          <img src="<%= image[:url] %>">
+        <% end %>
```
trotzig: I had thought about doing this before adding the count to the header. And after that it didn't feel like it was needed. I'm okay with this anyway. The `<hr>` removal was a good move.

lib/happo/public/happo-styles.css
```diff
@@ -24,7 +24,7 @@ body {
 .header {
   align-items: baseline;
   background-color: #ffffff;
-  border-bottom: 1px solid #cccccc;
+  box-shadow: 0 0 4px rgba(0, 0, 0, .3);
```
trotzig: 👍

lib/happo/views/diffs.erb
```diff
@@ -3,7 +3,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Happo diffs</title>
+    <title><%= Happo::Utils.page_title(diff_images, new_images) %></title>
```
trotzig: Neat!

lib/happo/views/diffs.erb
```diff
@@ -20,21 +20,35 @@
...
+          </a>
+        </h2>
+
         <% new_images.each do |image| %>
-          <h3>
-            <%= Rack::Utils.escape_html(image[:description]) %> @ <%= image[:viewport] %>
+          <h3 id="<%= Happo::Utils.image_slug(image) %>">
+            <a href="#<%= Happo::Utils.image_slug(image) %>">
+              <%= Rack::Utils.escape_html(image[:description]) %> @ <%= image[:viewport] %>
+            </a>
```
trotzig: It might be worth doing something about the code duplication here.

review-demo.rb
lencioni: I did, but I wanted it to be available when developing in the repo without having to install the gem and this seemed like the easiest way to make that possible.

trotzig: That makes sense. I've made good use of this thing since. Thanks for adding it!

Issue #61, opened 2015-10-31
Improve UI for reviewing diffs


trotzig:
Right now, we just inject the full diff image produced by diffux-core onto the page. It is sometimes hard to understand what the image is showing, especially for wide diffs. In the old Diffux project, we had a tool that allowed you to quickly switch between the before, after and diff image by simply hitting the `x` key. You could also click the tabs to switch. I think we should have something like that in the Diffux-CI project as well.

There are a few complexities here to consider:

- Most of the reviewing is done on the html page being uploaded to S3. This page needs (should) be fairly lightweight as we create it over and over again. We should probably also inject the javascript into the document as part of rendering it, so that we don't have to worry about uploading and accessing external assets. 

- I kind of want to use React, but it feels a bit heavy-handed to have to use a Babel/webpack/younameit process in order to generate the js to use on the page. We could use [in-browser es6 and jsx transformation](https://facebook.github.io/react/docs/tooling-integration.html#jsx) but that's not without its own set of complexities.

Any ideas here are welcome!

lencioni:
> I kind of want to use React

I'd also like to use React.

Here's an idea: what if we built a separate npm package that we get hosted on cdnjs.com or some other shared place. Then the JS we are adding to the page and uploading to S3 will be minimal for each set of diffs.

As far as using Babel, there is a [Babel gem](https://github.com/babel/ruby-babel-transpiler) that we could use to transpile our code pretty easily. We are using this for the style guide at Brigade and it was easy to set up and seems to be working well.

trotzig:
Sure. I hesitate a little bit about breaking out a separate project for this purpose (diffux is already scattered), but it makes sense. We can make it a general image-diff-review module that you can use for other purposes as well. But I guess the diff image format used by Diffux is very specific, though it is mostly optimized to support larger, full-page, diffs. Maybe we should just abandon that format and instead create diff images on the fly when you review? I remember seeing a technique using css to accomplish the overlay diff (can't find it now) and with such we could just save the baseline and candidate images. That should give a good performance boost too.

```jsx
<VisualDiffReview 
  before="path/to/baseline.png" 
  after="path/to/candidate.png"
/>
```

If we don't use the diff image from diffux-core, we could cut that dependency and just make a quicker pixel-by-pixel diff instead of the LCS thing. Trading precision for performance seems worth it here, doesn't it?

lencioni:
[...]

PR #119, opened 2016-08-28
Use React to render Diffs page

trotzig:
We're about to make the diff review experience better. To make future
enhancements easier, we're moving over the UI to React. This will allow
us to work inside an environment that we're all (Galooshi members)
familiar with.

I've been pushing back on introducing unnecessary complexity into this
project. We've had a fruitful discussion in #61 about possible ways to
move forward, and what I'm presenting here is basically a middle ground
between what @lencioni is suggesting and what I've been advocating for.

I'm making the page depend on three externally hosted scripts:
- React - so that we can write nice UIs
- ReactDOM - so that we can hook up those UIs to the DOM
- babel - so that we can use jsx, string templates, arrow functions and
  such.

The in-browser transpilation done by babel isn't the fastest, but I
don't see this as a bottleneck at this point.

If the application continues to grow, it might be worth introducing a
build step to the process. But at this point I don't think it's
necessary.

When I was porting over the image_slug method, I had some issue
reimplementing it in javascript. So I decided to delegate to
window.btoa() instead, which will simply base64 encode the string. This
fixes a potential naming collision issue, but at the cost of making the
URL less readable. Simplicity and robustness won this time, but there's
a chance we'll move over to more readable slugs at some later point in
time.

This change shouldn't result in any differences in the UI (except for
the already-mentioned slug change). Future commits will focus on
improving it.


lib/happo/public/HappoDiffs.jsx
```diff
@@ -0,0 +1,92 @@
+const PropTypes = React.PropTypes;
+
+const imageObjectStructure = {
+  description: PropTypes.string.isRequired,
+  viewport: PropTypes.string.isRequired,
+  url: PropTypes.string.isRequired,
+};
+
+function imageSlug(image) {
+  // TODO can we just use btoa() here?
```
lencioni: I don't understand this todo

lib/happo/views/diffs.erb
```diff
@@ -10,52 +10,27 @@
...
 
-        <% diff_images.each do |image| %>
-          <h3 id="<%= Happo::Utils.image_slug(image) %>">
-            <a class="anchored" href="#<%= Happo::Utils.image_slug(image) %>">
-              <%= Rack::Utils.escape_html(image[:description]) %> @ <%= image[:viewport] %>
-            </a>
-          </h3>
+    <script src="https://npmcdn.com/react@15.3.1/dist/react.min.js"></script>
+    <script src="https://npmcdn.com/react-dom@15.3.1/dist/react-dom.min.js"></script>
+    <script src="https://npmcdn.com/babel-core@5.8.38/browser.min.js"></script>
```
lencioni: I am a little worried about what happens if npmcdn is unavailable. It seems like it should be pretty reliable, but I'm pretty sure it is just one person's side-project. 

> I do my best to keep it running, but sometimes things go wrong. Sometimes there are network or provider issues outside my control. Sometimes abusive traffic temporarily affects response times. Sometimes I break things by doing something dumb, but I try not to.

https://npmcdn.com/#/about

I wonder how hard it would be to have a fallback in place.

lib/happo/views/diffs.erb
```diff
@@ -10,52 +10,27 @@
...
+  <body>
+    <script type="text/babel">
+      ReactDOM.render(
+        <HappoDiffs
+          newImages={<%= new_images.to_json %>}
+          diffImages={<%= diff_images.to_json %>}
+          pageTitle={<%= Happo::Utils.page_title(diff_images, new_images).to_json %>}
+          generatedAt={<%= Time.now.to_s.to_json %>}
+        />,
+        document.body
```
lencioni: I've read that people suggest not rendering directly into the body because it doesn't work very well with some things. It might be best to render into a div.

lencioni:
LGTM

lib/happo/public/HappoDiffs.jsx
trotzig: Oh, that shouldn't be there since I changed it. :)

lib/happo/views/diffs.erb
trotzig: We could have a fallback to another cdn hosted version. Falling back to a local file would mean always uploading the fallback for every serialized diffs page. 

This would work I think:
```js
window.React || document.write('<script src="fallback-url" />');
```

lib/happo/views/diffs.erb
trotzig: I see. That should be an easy fix.

trotzig:
I merged this on the command-line because I accidentally pushed a WIP that shouldn't be in the PR.

PR #120, opened 2016-08-29
Break apart diff image

trotzig:
This commit is a step in the direction of improving the diffs page.
Instead of generating a big diff image consisting of the before, after
and overlay, we instead keep all images separated.

We've been using the diff-lcs gem to construct the big image. This meant
that the diff contained gaps where the new snapshot had added or deleted
content. For big diffs, this was a good thing. But for small components,
the gaps often made the diffs harder to interpret. I'm simplifying the
overlayed image in this commit which means removing the gaps (and the
diff-lcs gem).

For now, the diffs UI is mostly kept the same as before. We're rendering
images left-to-right. Future commits will take advantage of the fact
that we now have individual images to work with.

Specs will be passing as part of this commit. I intend to follow up with
a simplification to the diff_in_percentage part, likely just replacing
it with a boolean (we're not really using the percentages anyhow).


trotzig:
Specs will break, but I decided to make it a PR anyway so that we can start the discussion. I'll push new commits to fix specs before merging this.

lencioni:
In your commit message you say

> Specs will be passing as part of this commit.

That's a typo, right?

lib/happo/public/HappoDiffs.jsx
```diff
@@ -3,24 +3,32 @@ const PropTypes = React.PropTypes;
...
+  previous: PropTypes.string,
+  current: PropTypes.string,
 };
 
 function imageSlug(image) {
   return btoa(image.description +  image.viewport);
 }
 
-function renderImage(image) {
+function renderImageHeading(image) {
```
lencioni: unrelated: These are very close to being stateless functional components. Might as well go all the way.

lib/happo/public/HappoDiffs.jsx
```diff
@@ -38,7 +46,12 @@ function renderDiffImages(diffImages) {
         </a>
       </h2>
 
-      {diffImages.map(renderImage)}
+      {diffImages.map((image, i) => (
+        <HappoDiff
+          key={i}
```
lencioni: I think the `image` would be a better key than the index.

trotzig:
haha, yes! I think I meant to say "Specs will not be passing".

lib/happo/public/HappoDiffs.jsx
trotzig: Oh, that's better.

lib/happo/public/HappoDiffs.jsx
```diff
@@ -56,11 +69,34 @@ function renderNewImages(newImages) {
...
+
+  render() {
+    const {
+      image,
+    } = this.props;
+
+    return (
+      <div>
+        {renderImageHeading(image)}
+        <div className='happo-diff__images'>
```
lencioni: nit: What do you think about PascalCasing the block on these to match the component? `HappoDiff__images`

lib/happo/public/HappoDiffs.jsx
trotzig: Yeah, you're right. 

Ninja-unrelated pingback: If you want to look into adding a build-step to the process, go ahead. I thought this file would remain pretty small, but it's obvious that it's about to grow too big pretty soon.

lib/happo/public/HappoDiffs.jsx
trotzig: I went back-and-forth... I'd be fine with that, though I'd prefer to use the new stuff you just open-sourced.

lencioni: Yeah, me too. Although we will likely want to bring in a build step for that since it requires a number of dependencies.

lib/happo/server.rb
```diff
@@ -27,25 +27,28 @@ def h(text)
...
 
-      diff_images = result_summary[:diff_examples].map do |example|
-        file_path = Happo::Utils.path_to(
-          example[:description],
-          example[:viewport],
-          'diff.png'
-        )
-        example[:url] = "/resource?file=#{ERB::Util.url_encode(file_path)}"
-        example
+      diff_images = result_summary[:diff_examples].map do |image|
```
lencioni: nit: I think `example` makes more sense as a variable name here than `image`.

lib/happo/server.rb
```diff
@@ -57,40 +60,46 @@ def h(text)
         {
           description: '<First> with "test"',
           viewport: 'small',
-          url: 'http://placehold.it/350x150',
+          previous: 'http://placehold.it/350x150',
+          diff: 'http://placehold.it/350x150',
+          current: 'http://placehold.it/350x150',
```
lencioni: It might be nice to use some different sizes for these, since that will likely happen in real life--particularly with height but also width sometimes too.

lib/happo/snapshot_comparer.rb
```diff
@@ -1,19 +1,24 @@
...
+    WHITE_OVERLAY = ChunkyPNG::Color.fade(WHITE, 1 - BASE_ALPHA)
+
     # @param png_before [ChunkyPNG::Image]
     # @param png_after  [ChunkyPNG::Image]
     def initialize(png_before, png_after)
       @png_after  = png_after
       @png_before = png_before
+
+      @diff_pixels  = {}
+      @faded_pixels = {}
```
lencioni: The name of this variable is confusing. The fact that these pixels are faded is an implementation detail of how we choose to represent them. I think it would be better to use a more semantic name, like we do for `diff_pixels`. How about `same_pixels` or `non_diff_pixels`?

lib/happo/snapshot_comparer.rb
```diff
@@ -27,65 +32,89 @@ def compare!
       # work.
       return no_diff if @png_before == @png_after
```
lencioni: unrelated: what do you think about moving `no_diff` up to a frozen constant to improve perf?

lib/happo/snapshot_comparer.rb
```diff
@@ -27,65 +32,89 @@ def compare!
...
+      max_height.times do |y|
+        max_width.times do |x|
+          pixel_before = @png_before.get_pixel(x, y)
+          pixel_after = @png_after.get_pixel(x, y)
+          number_of_different_pixels += 1 unless pixel_before == pixel_after
+          diff_image.set_pixel(x, y, faded_magenta_pixel(
+            pixel_after  || TRANSPARENT,
+            pixel_before || TRANSPARENT
+          ))
+        end
```
lencioni: I think we still want the diff-lcs algorithm here. We often have components where a new section is added and we don't want that blowing out the whole diff.

lib/happo/snapshot_comparer.rb
```diff
@@ -27,65 +32,89 @@ def compare!
...
+      number_of_different_pixels = 0
 
-      sdiff.each_with_index do |row, y|
-        # each row is a Diff::LCS::ContextChange instance
-        all_comparisons.each { |image| image.render_row(y, row) }
-        number_of_different_rows += 1 unless row.unchanged?
+      max_height.times do |y|
+        max_width.times do |x|
+          pixel_before = @png_before.get_pixel(x, y)
+          pixel_after = @png_after.get_pixel(x, y)
```
lencioni: If the images have different dimensions, do these methods always succeed? Do they just return `nil` or something?

lib/happo/snapshot_comparer.rb
```diff
@@ -27,65 +32,89 @@ def compare!
...
+    # Delta E in Lab colorspace, we probably don't need perceptual accuracy for
+    # this application, and it is nice to avoid the overhead of converting RGBA
+    # to Lab.
+    #
+    # @param pixel_after [Integer]
+    # @param pixel_before [Integer]
+    # @return [Float] number between 0 and 1 where 1 is completely different
+    #   and 0 is no difference
+    def pixel_diff_score(pixel_after, pixel_before)
+      ChunkyPNG::Color::euclidean_distance_rgba(pixel_after, pixel_before) /
```
lencioni: I wonder if it would be faster to do a simple equality check here first:

```rb
return 0 if pixel_after == pixel_before
```

lencioni:
Good start!

lib/happo/snapshot_comparer.rb
trotzig: Yeah, this is why I'm using the method and not `[x, y]` (which throws).

lib/happo/snapshot_comparer.rb
trotzig: Hrm, I'm not sure. Take a look at this:

-debatecard- with truncated best reasons_extra_large

We get those a ton, and it's pretty much impossible to spot the changes.

lib/happo/snapshot_comparer.rb
trotzig: Could be, I just copied this from overlayed.rb

lib/happo/snapshot_comparer.rb
lencioni: I think we could probably use both. The basic one that you have implemented in this PR can be pretty easily implemented in JS or CSS on the client side. The diff-lcs is very useful in some situations and is not as easily achievable in JS.

Although, maybe we should do all of these in JS on the client and check for any difference in the happo run. That would be a nice perf improvement and would bring us a step closer to rewriting the whole thing in JS. #81

lib/happo/snapshot_comparer.rb
lencioni: Yeah, I think it would improve perf a little. But I think we should just add it to chunkypng instead. https://github.com/wvanbergen/chunky_png/blob/cf9cf86894c4490421bb73263a8379ae4f2abfab/lib/chunky_png/color.rb#L716 I'll submit a PR over there.

lencioni: https://github.com/wvanbergen/chunky_png/pull/113

lib/happo/snapshot_comparer.rb
trotzig: > The basic one that you have implemented in this PR can be pretty easily implemented in JS or CSS on the client side.

Good point. 

> Although, maybe we should do all of these in JS on the client and check for any difference in the happo run.
Not sure I'm following/understanding how this can be accomplished...

lib/happo/snapshot_comparer.rb
lencioni: https://github.com/wvanbergen/oily_png/pull/17

lib/happo/snapshot_comparer.rb
lencioni: We would need to write some JS that reads in both images, performs the diff lcs algorithm, and then writes out a new image. I think we can probably use canvas for this, or generate a data URI or something. I don't have all the answers yet but I believe it would be possible.

trotzig:
I ended up force-pushing the commit to bring back lcs and the gutter etc.

lib/happo/server.rb
trotzig: Yeah, I kind of wanted to switch these out for "real" diffs at some point too. I think we could use some of the diffs produced by the brigade ci run.

lib/happo/public/HappoDiffs.jsx
```diff
@@ -24,78 +24,77 @@ function renderImageHeading(image) {
...
+        />
+      )}
     </div>
   );
 }
 
-const HappoDiff = React.createClass({
-  propTypes: {
-    image: PropTypes.shape(imageObjectStructure),
-  },
```
lencioni: You can still add propTypes to SFCs if you want.

lib/happo/public/HappoDiffs.jsx
```diff
@@ -24,78 +24,77 @@ function renderImageHeading(image) {
...
+        <img src={image.current} />
       </div>
-    );
-  }
-});
+    </div>
+  );
+}
 
 window.HappoDiffs = React.createClass({
```
lencioni: This could also be an SFC

lib/happo/public/HappoDiffs.jsx
```diff
@@ -81,20 +81,66 @@ function HappoNewImages({ images }) {
...
 
-function HappoDiff({ image }) {
-  return (
-    <div>
-      <HappoImageHeading
-        image={image}
-      />
-      <div className='happo-diff__images'>
-        <img src={image.previous} />
+const HappoDiff = React.createClass({
```
lencioni: It would be a little nicer to use classes here instead of the old createClass syntax.

lib/happo/public/HappoDiffs.jsx
```diff
@@ -81,20 +81,66 @@ function HappoNewImages({ images }) {
...
+    return {
+      selectedView: 'side-by-side',
+    };
+  },
+
+  _renderSelectedView() {
+    const { image } = this.props;
+    const { selectedView } = this.state;
+
+    if (selectedView === 'side-by-side') {
```
lencioni: These strings should be constants

lib/happo/public/HappoDiffs.jsx
```diff
@@ -81,20 +81,66 @@ function HappoNewImages({ images }) {
...
+    image: PropTypes.shape(imageObjectStructure),
+  },
+
+  getInitialState() {
+    return {
+      selectedView: 'side-by-side',
+    };
+  },
+
+  _renderSelectedView() {
```
lencioni: This should be a SFC

lib/happo/public/HappoDiffs.jsx
```diff
@@ -81,20 +81,66 @@ function HappoNewImages({ images }) {
...
+
+    return (
+      <div>
+        <HappoImageHeading
+          image={image}
+        />
+        <div className='happo-diff__buttons'>
+          {['side-by-side', 'diff'].map((view) => {
+            return (
+              <button
```
lencioni: this needs a `key`

lencioni:
LGTM

lib/happo/public/HappoDiffs.jsx
```diff
@@ -81,20 +81,66 @@ function HappoNewImages({ images }) {
...
-        image={image}
-      />
-      <div className='happo-diff__images'>
-        <img src={image.previous} />
+const HappoDiff = React.createClass({
+  propTypes: {
+    image: PropTypes.shape(imageObjectStructure),
+  },
+
+  getInitialState() {
```
lencioni: It would be nice to split this component into a container component that handles the state and a presentational component that renders the DOM.

Issue #121, opened 2016-08-29
Diffs page title has weird character in it


lencioni:
I just updated to Happo 2.3.0 and my first diffs upload has a weird character in its title.

> 2 diffs Â· Happo

![screen shot 2016-08-29 at 8 29 22 am](https://cloud.githubusercontent.com/assets/195534/18056938/d61f333c-6dc2-11e6-88e0-14c4d6f51e1e.png)

This upload had 2 diffs and no new examples.

trotzig:
I bet we need to explicitly set encoding somewhere. I can't find where though from the s3 docs (which are huuuuge).

lencioni:
I have verified that this has been fixed in v2.4.0.

PR #122, opened 2016-08-29
Add meta charset utf-8 to all pages

lencioni:
We are currently seeing a weird encoding issue on the uploaded diffs
page. I'm not sure if this will resolve that, but it is the right thing
to do and it might help.

Addresses #121

### DIFF
--- a/lib/happo/views/diffs.erb
+++ b/lib/happo/views/diffs.erb
@@ -3,6 +3,7 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <meta charset="utf-8" />
     <title><%= Happo::Utils.page_title(diff_images, new_images) %></title>
 
     <link rel="shortcut icon" href="<%= Happo::Utils.favicon_as_base64 %>" />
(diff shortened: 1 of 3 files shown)


lib/happo/views/diffs.erb
```diff
@@ -4,6 +4,7 @@
 <html>
   <head>
     <title><%= Happo::Utils.page_title(diff_images, new_images) %></title>
+    <meta charset="utf-8" />
```
trotzig: I think it would be better to put the meta tag above the title.

trotzig:
Did you try dynamically adding this to your broken diff page? Maybe that won't have any effect though, I don't really understand how you can switch encoding mid-document...

lib/happo/views/diffs.erb
lencioni: Good call. Done!

trotzig:
👍

lencioni:
I don't think that would have any affect. But I'll give it a shot.

lencioni:
Dynamically adding it did nothing.

PR #124, opened 2016-08-30
Set up eslint

lencioni:
There are still a few rule violations, but this is a good step in the right direction.

### DIFF
--- a/lib/happo/public/HappoDiffs.jsx
+++ b/lib/happo/public/HappoDiffs.jsx
@@ -81,6 +81,23 @@ function HappoNewImages({ images }) {
   );
 }
 
+function SelectedView({ image, selectedView }) {
+  if (selectedView === 'side-by-side') {
+    return (
+      <div>
+        <img src={image.previous} />
+        <img src={image.current} />
+      </div>
+    );
+  }
+
+  if (selectedView === 'diff') {
+    return (
+      <img src={image.diff} />
+    );
+  }
+};
+
 const HappoDiff = React.createClass({
   propTypes: {
     image: PropTypes.shape(imageObjectStructure),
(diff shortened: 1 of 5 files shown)


lib/happo/public/HappoDiffs.jsx
```diff
@@ -80,10 +90,34 @@ function HappoNewImages({ images }) {
     </div>
   );
 }
+HappoNewImages.propTypes = {
+  images: PropTypes.arrayOf(PropTypes.shape(imageShape)),
+};
+
+function SelectedView({ image, selectedView }) {
```
trotzig: I've been using "Happo" as a prefix to all of these. Don't exactly know why, but it might be worth being consistent.

lib/happo/public/happo-runner.js
```diff
@@ -6,11 +13,11 @@ window.happo = {
   currentRenderedElement: undefined,
   errors: [],
 
-  define: function(description, func, options) {
+  define: function define(description, func, options) {
```
trotzig: It looks a little odd having `define` duplicated here.

trotzig:
looks good to me!

lib/happo/public/happo-runner.js
lencioni: Indeed, but it is better because these function names will show up in stack traces and profiling. We could make this nicer if we used object shorthand syntax, but we'd need to transpile this then.

lib/happo/public/HappoDiffs.jsx
lencioni: I think we can drop the Happo prefix across the board.

lib/happo/public/happo-runner.js
trotzig: Are you sure that doesn't just work? The babel in-browser transpiler has surprised me before.

lencioni: Are we running the babel in-browser transpiler on the happo runs?

trotzig: Oh, you're right. No we don't.

PR #126, opened 2016-08-31
View on top

trotzig:


trotzig:
I forgot to PRize this last night. I can rebase on top of your PR later @lencioni.

lib/happo/public/HappoDiffs.jsx
```diff
@@ -94,21 +100,91 @@ HappoNewImages.propTypes = {
...
+      cursorLeftPercent: 50, // Start in the center
+    };
+  },
+
+  _handleMouseMove(event) {
+    const leftPx = event.pageX - event.target.offsetLeft;
+    const leftPercent = leftPx / event.target.offsetWidth * 100;
+
+    console.log(event.target)
+    console.log(leftPx, event.target.offsetWidth, leftPercent);
```
lencioni: console.log

lib/happo/public/HappoDiffs.jsx
```diff
@@ -94,21 +100,91 @@ HappoNewImages.propTypes = {
...
+      previous,
+      current,
+    } = this.props;
+
+    const { cursorLeftPercent } = this.state;
+    const clipPath =
+      `polygon(${cursorLeftPercent}% 0, 100% 0, 100% 100%, ${cursorLeftPercent}% 100%)`;
+
+    return (
+      <div
```
lencioni: It would be nice to split this component into a controller component and a presentational component.

lib/happo/public/HappoDiffs.jsx
```diff
@@ -94,21 +100,91 @@ HappoNewImages.propTypes = {
   images: PropTypes.arrayOf(PropTypes.shape(imageShape)),
 };
 
+const Swiper = React.createClass({
```
lencioni: This should be class syntax

lib/happo/public/HappoDiffs.jsx
```diff
@@ -94,21 +100,91 @@ HappoNewImages.propTypes = {
...
+      current,
+    } = this.props;
+
+    const { cursorLeftPercent } = this.state;
+    const clipPath =
+      `polygon(${cursorLeftPercent}% 0, 100% 0, 100% 100%, ${cursorLeftPercent}% 100%)`;
+
+    return (
+      <div
+        className='happo-swiper'
```
lencioni: className should be `Swiper`

lib/happo/public/HappoDiffs.jsx
```diff
@@ -94,21 +100,91 @@ HappoNewImages.propTypes = {
...
+        style={{
+          backgroundImage: `url(${previous})`,
+        }}
+        onMouseMove={this._handleMouseMove}
+      >
+        <img
+          className='happo-swiper__image'
+          style={{
+            WebkitClipPath: clipPath,
```
lencioni: Is it worth trying to use a different solution for broader browser support?

lib/happo/public/HappoDiffs.jsx
```diff
@@ -94,21 +100,91 @@ HappoNewImages.propTypes = {
...
+            clipPath: clipPath,
+          }}
+          src={current}
+        />
+        <div
+          className='happo-swiper__cursor'
+          style={{
+            left: `${cursorLeftPercent}%`,
```
lencioni: Would it be smoother to use a translation instead of left here?

lib/happo/public/HappoDiffs.jsx
```diff
@@ -94,21 +100,91 @@ HappoNewImages.propTypes = {
...
+  }
+});
+
 function SelectedView({ image, selectedView }) {
-  if (selectedView === 'side-by-side') {
+  if (selectedView === VIEWS.sideBySide) {
     return (
       <div>
         <img src={image.previous} />
+        {' '}
```
lencioni: It's funny that we made some of the same decisions, like this one.

lib/happo/public/HappoDiffs.jsx
```diff
@@ -94,21 +100,91 @@ HappoNewImages.propTypes = {
...
+
+    const { cursorLeftPercent } = this.state;
+    const clipPath =
+      `polygon(${cursorLeftPercent}% 0, 100% 0, 100% 100%, ${cursorLeftPercent}% 100%)`;
+
+    return (
+      <div
+        className='happo-swiper'
+        style={{
+          backgroundImage: `url(${previous})`,
```
lencioni: If the previous image is bigger than the current, how will this look? It might be good to modify some of the example images to showcase this scenario (and the reverse).

lib/happo/public/HappoDiffs.jsx
trotzig: ooops. Removed now.

lib/happo/public/HappoDiffs.jsx
trotzig: Can you explain this more? I understand how that would make happoing the component easier. But it comes with a pretty weird dependency between the controller and the child, where I feel like I'm just boilerplate-passing around stuff back and forth.

lib/happo/public/HappoDiffs.jsx
trotzig: done

lib/happo/public/HappoDiffs.jsx
trotzig: done.

lib/happo/public/HappoDiffs.jsx
trotzig: I'll look into this. I added Moz to the mix as well, and it's working fine. I'm not too worried about IE support at this point.

lib/happo/public/HappoDiffs.jsx
trotzig: This is smooth enough if you ask me.

lib/happo/public/HappoDiffs.jsx
trotzig: Great minds think.

lib/happo/public/HappoDiffs.jsx
trotzig: Good point. I mentioned this limitation in the commit message.

trotzig:
I'm planning on making a release out of this tomorrow.

lib/happo/public/HappoDiffs.jsx
lencioni: I don't see it as a hard rule, but I generally find that this pattern makes an app more understandable and testable. It's kinda like what Dan outlines here: https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0

lib/happo/public/HappoDiffs.jsx
lencioni: Does this work in Safari? MDN doesn't know.

If you find out, want to update MDN? https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path

lib/happo/public/happo-styles.css
```diff
@@ -107,3 +107,26 @@ body {
...
+
+.SideBySide__caption {
+  background-color: #666666;
+  bottom: 0;
+  color: #ffffff;
+  font-size: 10px;
+  line-height: 1;
+  opacity: 0;
+  padding: 3px 6px;
+  position: absolute;
```
lencioni: Maybe add `pointer-events: none`?

lib/happo/public/happo-styles.css
```diff
@@ -107,3 +107,26 @@ body {
...
+  right: 0;
+  transition: opacity .2s;
+}
+
+.SideBySide__image:hover .SideBySide__caption {
```
lencioni: I'm not super keen on showing things only on hover, but maybe this is okay. Did you consider using a `title` attribute instead?

lib/happo/public/HappoDiffs.jsx
```diff
@@ -114,6 +115,69 @@ NewImages.propTypes = {
...
+    ];
+    const clipPath = `polygon(${polygonPoints.join(',')})`;
+
+    return (
+      <div
+        className='Swiper'
+        style={{
+          backgroundImage: `url(${previous})`,
+        }}
+        onMouseMove={this.handleMouseMove}
```
lencioni: What do you think about making this so it only moves when the mouse button is down?

lib/happo/public/happo-styles.css
trotzig: I guess, but I didn't see a need for it now.

lib/happo/public/happo-styles.css
trotzig: I don't know why i didn't think about that. That's so much simpler. I might change this later on.

lib/happo/public/HappoDiffs.jsx
trotzig: That would require a UI with better affordance I think. But I'm open to this.

PR #127, opened 2016-09-01
Make image swiping work for all image sizes

lencioni:
Our first pass at this didn't show the entire image if the previous
image was larger than the current image. I decided to fix this by
taking an alternate approach that uses containers with hidden
overflow. This gives us more control over how these elements behave.

Since this technique uses translateX, which uses percentages based off
the element instead of the container, I needed to switch the cursor
left position from being a percentage to being an absolute number.

### DIFF
--- a/lib/happo/public/HappoDiffs.jsx
+++ b/lib/happo/public/HappoDiffs.jsx
@@ -116,58 +116,116 @@ NewImages.propTypes = {
   images: PropTypes.arrayOf(PropTypes.shape(imageShape)).isRequired,
 };
 
+function maxImageSize(...imageUrls) {
+  const dimensions = {};
+
+  return new Promise((resolve, reject) => {
+    imageUrls.forEach((url, i) => {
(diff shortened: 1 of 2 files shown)


lib/happo/public/HappoDiffs.jsx
```diff
@@ -116,58 +116,116 @@ NewImages.propTypes = {
...
+    this.updateSize(this.props)
+      .then(({ width }) => {
+        // Start in the center
+        this.setState({ cursorLeft: width / 2 });
+      });
+  }
 
+  componentWillReceiveProps(nextProps) {
+    this.updateSize(nextProps);
+  }
```
trotzig: Will we ever send new props?

lencioni: We don't right now, but I wanted this to work regardless and it was easy enough to do.

trotzig:
Looks good!

PR #144, opened 2016-09-11
Lcs in browser

trotzig:


trotzig:
I made this branch into a PR so that it's easier to keep the discussion going. 

The whole PR is still WIP, so I don't think we should merge yet. I unwipped the last commit in the branch by making it draw a simple overlay image (no lcs diffing yet).

lencioni:
`JSON.stringify` has been aggressively optimized by browsers over the past few years, since it is so widely used. Hard to top it.

trotzig:
I thought those comments ^^ would be attached to the commit I posted them on. Apparently not... I find the reviewing UI confusing...

lencioni:
I think those are more like "cover pages" for your whole review, which is on a PR basis and can span multiple commits.

js/src/maxImageSize.js
```diff
@@ -2,28 +2,29 @@ export default function maxImageSize(...imageUrls) {
...
+        const maxWidth = Math.max(...values.map(value => value.width));
+        const maxHeight = Math.max(...values.map(value => value.height));
+        resolve({ width: maxWidth, height: maxHeight });
+      }
+    }
 
+    imageUrls.forEach((url, i) => {
+      const image = new Image();
+      image.onerror = handleImageError;
+      image.onload = handleImageLoad.bind(image, i);
```
trotzig: You could probably move this functions outside of the Promise too, with some cleverness of making sure that they're still attached to the original promise (passing in `resolve` as a `.bind()` perhaps?).

lencioni: I'm not sure I follow, and I also don't see why we would care to do that.

trotzig: Instead of creating the functions inside the promise we could define them outside. I don't think it matters though.

js/src/LCSDiff.jsx
```diff
@@ -68,14 +81,31 @@ export default class LCSDiff extends React.Component {
...
-      <canvas
-        ref={(node) => { this.canvas = node; }}
-        width={width}
-        height={height}
-      />
+      <div>
+        {progress === 0 &&
+          <ReactWaypoint
+            onEnter={this.initialize}
+            scrollableAncestor={window}
```
lencioni: Should we add an offset so we can start loading things before they scroll into the viewport? I think that would make the interface feel better.

js/src/workers/ComputeAndInjectDiffsWorker.js
```diff
@@ -89,13 +89,17 @@ function computeAndInjectDiffs({ previousData, currentData }) {
   const currentImageData = imageTo2DArray(
     currentData, maxWidth - currentData.width);
 
+  self.postMessage({ progress: 30 });
```
lencioni: If the diff takes disproportionately longer than the other operations as the image size goes up, it might be nice to come up with a quick formula to make these progress numbers more accurate.

package.json
```diff
@@ -33,6 +33,9 @@
   },
   "dependencies": {
     "adiff": "^0.2.13",
+    "react": "^15.3.1",
```
lencioni: Should we switch to using the installed package instead of using it as an external?

lencioni:
FYI, I noticed Travis runs are really slow on this branch (~15min).

js/src/LCSDiff.jsx
trotzig: Yeah, good idea. I'll do that (unless you beat me to it).

js/src/workers/ComputeAndInjectDiffsWorker.js
trotzig: Yeah, not sure if this is worth the effort. I think I'm just going to show a "spinner" type indicator (not necessarily a spinning thing, but you get the point).

package.json
trotzig: I don't think that's a good idea. It adds a significant weight to the serialized diff page.

spec/happo_spec.rb
```diff
@@ -102,7 +102,6 @@ def snapshot_file_exists?(description, size, file_name)
         ],
         diff_examples: [],
         okay_examples: [],
-        height: 20,
```
trotzig: Thanks!

Issue #78, opened 2016-07-07
Add example for working with React


lencioni:
All of the examples show how to set this up for generic JavaScript, which is great. However, if you want to get this set up for a project that uses React, there might be a couple of missing pieces. It would be nice if there was a clear React example or some documentation for this.

oliviertassinari:
For what that worth, I have been [adding an example](https://github.com/argos-ci/argos/tree/master/examples/with-happo) with how to use Happo with React, using Webpack under the hood to load components. I'm not sure it's the right way to do it. I have used the [vrtest](https://github.com/nathanmarks/vrtest) set up of [Material-UI](https://github.com/callemall/material-ui/tree/next/test) as a starting point.

PR #149, opened 2016-09-13
Get root elements

trotzig:


lencioni:
This also partly addresses #78.

PR #150, opened 2016-09-13
Don't assume elements start at top left corner

trotzig:
Due to how we initialized the box at the top left corner, all snapshots
always started at the top left. This isn't great if you have a component
rendering at the right or bottom edge of the screen.

In the Brigade codebase, we have a <Toast> component that renders at the
bottom of the screen. Before this change, the snapshot would include a
large gray box above the component. After this change, only the
component is included in the snapshot.

### DIFF
--- a/lib/happo/public/happo-runner.js
+++ b/lib/happo/public/happo-runner.js
@@ -182,6 +182,16 @@ window.happo = {
     }
   },
 
+  /**
+   * Wrapper around Math.min to handle undefined values.
+   */
+  min: function min(a, b) {
+    if (a === undefined) {
+      return b;
+    }
+    return Math.min(a, b);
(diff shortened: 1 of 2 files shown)


lib/happo/public/happo-runner.js
```diff
@@ -221,9 +231,9 @@ window.happo = {
     // Set up the initial object that we will mutate in our recursive function.
     var box = {
       bottom: 0,
-      left: 0,
+      left: undefined,
```
lencioni: Instead of adding the wrapper, you could probably use `Infinity` for these values.

trotzig: Good point, I hadn't thought of that.

PR #158, opened 2016-09-27
Replace adiff with our own, iterative, LCS implementation

trotzig:
The adiff package served us well, but it was ultimately not the right
tool for the job:

- It couldn't handle large arrays well (maximum call stack size exceeded)
- It wasn't as efficient as we wanted
- It would compute results that we didn't need

Most of those ^^ are covered in [1].

I've been exploring rolling our own LCS implementation for a while, and
this commit is the result of that. Instead of using recursion (as adiff
is doing), we're taking an iterative approach. By using memoization and
a traceback approach, we can carefully inject placeholders in the
before and after array to make the two align with each other. Consider
this example:

  A = 'ABCD'
  B = 'GAFFC'

We can see that the LCS of these two strings is 'AC'. Now, if we inject
placeholders (the +'es below) in a few places we can make the A and C
characters align:

  A = '+AB+CD'
  B = 'GAFFC+'

When the arrays align, we know that they're of the same size. We can
then feed them along to the ImageDiff worker comparing them pixel by
pixel.

To be able to test the algorithm, I've added a Jest/Jasmine test. For
that to work, I had to add a few package dependencies, following the
guide on the Jest homepage [2].

While developing the LCS implementation, I found this to be a good
resource:
http://algorithms.tutorialhorizon.com/dynamic-programming-longest-common-subsequence/
The `longestCommonSubsequence` function is basically my interpretation
of the (Java) example on that page.

[1]: https://github.com/dominictarr/adiff/issues/7
[2]: https://facebook.github.io/jest/


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Command-line tool to visually diff JavaScript components",
   "scripts": {
     "eslint": "eslint --ext .js,.jsx",
-    "test": "npm run --silent eslint -- .",
+    "test": "jest",
```
lencioni: Oops, it looks like this change makes it so eslint is no longer running in CI!

trotzig:
Crap, this was a little sloppy. I'll put it back.

package.json
trotzig: I don't think I understand how that is set up. Is this done automagically by travis? I don't see it referenced in `.travis.yml`. 

I'll put it back.

lencioni:
Yeah, for node jobs, Travis will run npm test by default.

PR #163, opened 2016-10-08
Wait for images to load

### LINKED ISSUE #162: Height of snapshot is wrong when using images with dynamic height
We compute a rectangle to crop out the relevant parts from the full-screen screenshot that Selenium produces. This rectangle is computed immediately after the component is rendered. The screenshot will wait for images to load. If images don't have a set height (and in some cases, width) the snapshot will be cropped according to the original rectangle (before image was loaded). 

Here's an example:
http://s3.amazonaws.com/diffux_ci-diffs/585a9297-0df1-44ab-847a-8a5fbc70d72b/index.html

In a way, this is good to know about a component. Because users are most likely experiencing a bouncy page load. But I think happo should do better here.

trotzig:
The goals of this PR are two-fold:

A: fix #162 by loading images before the screenshot is taken
~~B: re-implement [the screenshotting function from Selenium's firefox driver]~~(https://github.com/SeleniumHQ/selenium/blob/ceaf3da79542024becdda5953059dfbb96fb3a90/javascript/firefox-driver/js/screenshooter.js#L23) and only grab the part of the screen we're interested in (hence the "No crop" name of the PR). 

EDIT: I ended up only including the image wait in this commit. Screenshot improvements will be done in another PR (as suggested by @lencioni).


trotzig:
I'm not done here, but I wanted to get this out so that you can feedback @lencioni.

js/src/HappoRunner.js
```diff
@@ -183,73 +153,15 @@ window.happo = {
...
-    box.top = Math.max(box.top, 0);
-
-    box.width = box.right - box.left;
-    box.height = box.bottom - box.top;
-
-    return box;
-  },
-
   processExample(currentExample) {
+    const rootNodes = this.getRootNodes();
```
lencioni: Does `getRootNodes` need to be on `this`? Are there other methods that can be moved off in addition to the ones you've already done?

js/src/HappoRunner.js
```diff
@@ -127,14 +128,14 @@ window.happo = {
           // The function returned a promise, so we need to wait for it to
           // resolve before proceeding.
           result.then(() => {
-            doneFunc(this.processExample(currentExample));
+            this.processExample(currentExample).then(doneFunc);
```
lencioni: I wonder if this change and the next one are what caused the spec failures.

js/src/HappoRunner.js
```diff
@@ -153,26 +154,33 @@ window.happo = {
...
+
+          resolve({
+            description: currentExample.description,
+            height,
+            left,
+            top,
+            width,
+          });
+        } catch (error) {
+          resolve(handleError(currentExample, error));
```
lencioni: Should this be `reject`?

trotzig: I guess, although we end up calling the `doneFunc` with an object that the ruby code inspects. This used to be a return so I just made it resolve.

js/src/HappoRunner.js
trotzig: `getRootNodes` is a function that you can override as a user. We should probably look into a better way of specifying that, but at this point we at least need to call the method. We could extract the default into an anonymous function but I don't know what that would give us.

js/src/HappoRunner.js
```diff
@@ -281,52 +271,57 @@ window.happo = {
...
+  const example = decodeURIComponent(matches[1]);
+  window.happo.renderExample(example, () => {});
 });
 
 // We need to redefine a few global functions that halt execution. Without this,
 // there's a chance that the Ruby code can't communicate with the browser.
-window.alert = function alert(message) {
-  console.log('`window.alert` called', message);
+window.alert = (message) => {
+  console.log('`window.alert` called', message); // eslint-disable-line
```
trotzig: Whoops, didn't mean to do this. Thanks!

js/src/HappoRunner.js
trotzig: Very likely. I'll look into it.

trotzig:
Okay, I tried the `ctx.drawWindow` function, and I believe it was added in a fairly recent Firefox version. It's not working in 47.0.1, which is the latest version I know we can run with Selenium. 😞

trotzig:
Alright, I think we're in a better spot now. This ended up being a lot more complicated than I thought (not the waiting for images part, but the travis build woes).

.travis.yml
```diff
@@ -9,9 +9,17 @@ rvm:
   - 2.1
   - 2.2
 
+env:
+  - TRAVIS_NODE_VERSION="6"
```
lencioni: This will be easier once we finish rewriting in JS.

.travis.yml
```diff
@@ -9,9 +9,17 @@ rvm:
...
+
 before_install:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
-  - "npm install && npm run build"
 
-script: bundle exec ruby headless_rspec.rb
+install:
+  - 'rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install $TRAVIS_NODE_VERSION'
```
lencioni: Hey, this looks kinda familiar! https://github.com/creationix/nvm/pull/1179

js/src/waitForImagesToRender.js
```diff
@@ -9,7 +9,7 @@ function waitForImageToLoad(url) {
 
 export default function waitForImagesToRender() {
   return new Promise((resolve, reject) => {
-    const promises = Array.from(document.querySelectorAll('img'))
+    const promises = Array.prototype.slice.call(document.querySelectorAll('img'))
```
lencioni: Did you consider this?

```js
const promises = [...document.querySelectorAll('img')]
```

.travis.yml
trotzig: lol!

js/src/waitForImagesToRender.js
trotzig: I didn't. Oh well, I think the slice will do for now.

PR #167, opened 2016-12-07
Add badges to readme

lencioni:
Nobody will take us seriously unless we have all of these badges.


trotzig:
Lol 😄 

Personally I'm not a big fan of these (I think they're mostly just noise). But I'm perfectly okay with having them though.

trotzig:
I'll let you merge however, I don't want any of my fingerprints on commits like these... ;)

PR #175, opened 2016-12-23
Wait for bg images

trotzig:


trotzig:
cc @lelandrichardson

trotzig:
Spec failures are legit. They're passing for me locally but that's because I'm on a newer firefox version. I'll update later on, ran out of time now. 

```
document.body.querySelectorAll(...).forEach is not a function
```

example-project/css/styles.css
```diff
@@ -14,3 +14,9 @@
   display: inline-block;
 
 }
+
+.big-background {
+  height: 200px;
+  width: 600px;
+  background-image: url('https://upload.wikimedia.org/wikipedia/commons/9/92/Big_Sur_Coast_California.JPG')
```
lencioni: I'm not sure it is a good idea to rely on an external resource in specs.

src/waitForImagesToRender.js
```diff
@@ -14,6 +14,16 @@ export default function waitForImagesToRender() {
       .filter(Boolean)
       .map(url => waitForImageToLoad(url));
 
+    document.body.querySelectorAll('*').forEach((element) => {
+      const computedStyle = window.getComputedStyle(element);
+      const bgImage = computedStyle.getPropertyValue('background-image');
+      if (bgImage && bgImage.startsWith('url(')) {
```
lencioni: This doesn't look like it will work if using multiple background images. https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Background_and_Borders/Using_CSS_multiple_backgrounds

src/waitForImagesToRender.js
```diff
@@ -14,6 +14,16 @@ export default function waitForImagesToRender() {
...
+        const url = computedStyle.getPropertyValue('background-image')
+          .replace(/^url\(['"]/, '').replace(/['"]?\)$/, '');
```
lencioni: This replace would be a bit better if it used a backreference.

Something like
```js
replace(/^url\((['"])(.+?)\1.*/, '$2')
```

example-project/css/styles.css
trotzig: This is in the example-project, so technically not a spec.

src/waitForImagesToRender.js
trotzig: Ah, good catch. I'll see if I have to take that into account as well.

PR #177, opened 2017-01-04
Pre cropped screenshots

trotzig:


trotzig:
@lencioni: Can you have a look at these changes if you have a moment? I still have test failures to fix which I won't have time to do until later during the week. So no rush.

trotzig:
I also need to update the README a little I believe.

src/server/checkBrowserVersion.js
```diff
@@ -0,0 +1,37 @@
...
+    new firefox.Binary().locate().then((pathToExecutable) => {
+      childProcess.exec(`${pathToExecutable} --version`, (error, stdout) => {
+        if (error) {
+          reject(new Error(`Failed to check Firefox version: ${error}`));
+          return;
+        }
+        const match = stdout.match(FIREFOX_VERSION_MATCHER);
+        if (!match) {
+          reject(new Error(`Failed to parse Firefox version string "${stdout}"`));
+        } else if (parseFloat(match[1]) < MINIMUM_FIREFOX_VERSION) {
```
lencioni: I am currently running Firefox 47.0.1. I think we should probably use something that can compare semver numbers to each other instead of just parseFloat.

src/server/checkBrowserVersion.js
```diff
@@ -0,0 +1,37 @@
+const childProcess = require('child_process');
+
+const firefox = require('selenium-webdriver/firefox');
+
+const config = require('./config');
+
+const MINIMUM_FIREFOX_VERSION = 50.0;
+const FIREFOX_VERSION_MATCHER = /Mozilla Firefox ([0-9.]+)/;
```
lencioni: Some Firefox versions end in `esr` (e.g. `45.6.0esr`). It would be nice if the error message we generate includes this detail for maximum clarity.

src/HappoRunner.js
```diff
@@ -170,6 +170,15 @@ window.happo = {
             width,
           } = getFullRect(rootNodes);
 
+          const overlay = document.createElement('div');
```
lencioni: It might be nice to have a comment here explaining why we are creating an overlay. Or perhaps a more descriptive variable name than `overlay`. Maybe `screenshotOverlay` or `screenshotBox`?

src/HappoRunner.js
```diff
@@ -170,6 +170,15 @@ window.happo = {
...
+          overlay.style.position = 'absolute';
+          overlay.style.left = `${left}px`;
+          overlay.style.top = `${top}px`;
+          overlay.style.width = `${width}px`;
+          overlay.style.height = `${height}px`;
+          overlay.setAttribute('id', 'happo-screenshot-overlay');
```
lencioni: This should probably be some sort of shared constant.

src/server/addGeckoDriverToPath.js
```diff
@@ -0,0 +1,10 @@
+const path = require('path');
+
+const { platform, arch } = process;
+
+const geckodriverFolder = path.join(__dirname,
+  `../../geckodriver/${platform}-${arch}/`);
+
+process.env.PATH += path.delimiter + geckodriverFolder;
+
+module.exports = geckodriverFolder;
```
lencioni: This export doesn't seem to be used. Remove it?

src/server/closeDriver.js
```diff
@@ -0,0 +1,36 @@
...
+
+// Works around an issue in marionette causing ghost firefox instances to remain
+// running.
+//
+// https://bugzilla.mozilla.org/show_bug.cgi?id=1310992
+// https://github.com/mozilla/geckodriver/issues/285
+module.exports = function closeDriver(driver) {
+  return new Promise((resolve, reject) => {
+    driver.close();
+    psNode.lookup({
```
lencioni: Is there a way to get the pid from the `driver` object instead of doing this lookup?

src/server/checkBrowserVersion.js
trotzig: We can do that when we know we need to, right? For now, this should be enough I think.

src/server/checkBrowserVersion.js
trotzig: It looks like we do, don't we? 

```
reject(new Error(`Failed to parse Firefox version string "${stdout}"`));
```

src/HappoRunner.js
trotzig: Good call. 90d68b57

src/server/addGeckoDriverToPath.js
trotzig: Thanks! 39dd22a

src/server/closeDriver.js
trotzig: Not that I can see. This is what I see it having: 

```
Driver {
  flow_:
   ControlFlow {
     propagateUnhandledRejections_: true,
     activeQueue_: null,
     taskQueues_: Set {},
     shutdownTask_: null,
     hold_: null },
  session_:
   ManagedPromise {
     flow_:
      ControlFlow {
        propagateUnhandledRejections_: true,
        activeQueue_: null,
        taskQueues_: Set {},
        shutdownTask_: null,
        hold_: null },
     stack_: null,
     parent_: null,
     callbacks_: null,
     state_: 'fulfilled',
     handled_: true,
     value_: Session { id_: 'ac1a9200-3bd4-ca42-a2af-b711fe062e05', caps_: [Object] },
     queue_:
      TaskQueue {
        name_: 'TaskQueue::23',
        flow_: [Object],
        tasks_: [],
        interrupts_: null,
        pending_: null,
        subQ_: null,
        state_: 'finished',
        unhandledRejections_: Set {} } },
  executor_:
   Executor {
     w3c: true,
     customCommands_: Map { 'getContext' => [Object], 'setContext' => [Object] },
     log_:
      Logger {
        name_: 'webdriver.http.Executor',
        level_: null,
        parent_: [Object],
        handlers_: null } },
  fileDetector_: null,
  onQuit_: [Function: onQuit] }
```

src/server/checkBrowserVersion.js
lencioni: Oh I was thinking of this line:

```
`You are using ${match[1]}`));
```

trotzig: Ah - of course. yagni?

lencioni: As it is written, people will potentially see a misleading error message, which seems like it should be fixed.

lencioni: Alternatively, you could make the error message less useful by not including the current version. That would be a strict improvement over giving potentially incorrect information.

trotzig: I made it print the full string instead: e16e3110

lencioni:
Maybe try updating to Jest 18?

trotzig:
Good call. I'll try that next.

trotzig:
I found [one issue](https://github.com/Galooshi/happo/issues/180) that we should fix. But I won't let that block shipping v4.

oliviertassinari:
I have been trying to port the `happo-target-firefox` to work with a docker bases selenium solution.
Didn't succeed as you said in the description, selenium does not support it.
That's a pretty nice finding!

PR #179, opened 2017-01-06
Add support for React Native

lelandrichardson:
to: @lencioni @trotzig 

This PR is a work in progress, and so far only works for ios (but is almost ready for android). It's best viewed by commit, as it includes a lot of RN template code in the first 2 commits.

Ultimately, I think it makes sense for us to implement this as a plugin system or something, so this PR will likely never get merged, but I thought it would be a good way for us to start the discussion.

I haven't made much of an attempt to share code, though I did refactor some methods into modules so they could be shared across implementations.

**How it works:**

running `happo run-ios` does the following things:

1. Starts a websockets server that the ios app will use to communicate between the two processes
2. Starts appium, which then launches the simulator with the happo app
3. The happo app starts, and requests an RN bundle from `localhost:8081/happo.bundle` which is assuming that the consumer of happo has the RN packager turned on (to default port of 8081) with a `happo.js` entry file at the root.
4. The `happo.js` file needs to at some point require the `happo/native` js module, which will register itself with the happo server, and the happo app.
5. Consumer of happo will register happo stories with `import StoryManager from 'happo/native';`
6. At startup, happo will send the list of story names over websockets to happo, then render them one by one, taking a snapshot of each, and saving to a temp file.  Temp files + story names will be sent across websockets to the happo server.
7. Happo server diffs files appropriately and does its thing


package.json
```diff
@@ -55,7 +58,12 @@
     "mkdirp": "^0.5.1",
     "png-crop": "0.0.1",
     "pngjs": "^2.3.1",
+    "react": "15.4.1",
```
lencioni: Should we remove this from devDepedencies at the same time? I suppose it won't matter once we actually do the plugin stuff we talked about.

src/server/RunResult.js
```diff
@@ -0,0 +1,29 @@
...
+    this.diffImages = [];
+  }
+
+  add({
+    result,
+    description,
+    height,
+    viewportName,
+  }) {
+    if (result === 'equal') {
```
lencioni: We should really make these constants at some point

src/server/areImagesEqual.js
```diff
@@ -0,0 +1,16 @@
+
+module.exports = function areImagesEqual(a, b) {
+  if (a.height !== b.height) {
+    return false;
+  }
+  if (a.width !== b.width) {
+    return false;
+  }
+  const len = a.data.length;
+  for (let i = 0; i < len; i += 1) {
```
lencioni: I know you just moved this function, but we never expect any transparency in our images, so you can make this faster by incrementing by +4 every step, and only checking the first three bits (r, g, b) and skipping the fourth (a).

lencioni: Although, I wonder if we can figure out how to take screenshots of things without a page background color so that there is transparency. That would make it possible to differentiate between a white background that a component adds and transparent background on the component.

src/server/runIOSVisualDiffs.js
```diff
@@ -0,0 +1,108 @@
...
+  const viewportName = 'ios';
+  return new Promise((resolve) => {
+    setupWebsockets(
+      (/* name, uri */) => {
+        // do nothing
+      },
+      (snapshots) => {
+        resolve(snapshots);
+      }
+    ).then(initializeIOSdriver);
```
lencioni: We have the initializeWebdriver call in the cli script, but you put initializeIOSdriver inside of runIOSVisualDiffs. I don't think it matters too much where it goes either way, but I think it would be good to have consistency.

src/server/runIOSVisualDiffs.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+  });
+}
+
+module.exports = function runIOSVisualDiffs() {
+  const viewportName = 'ios';
+  return new Promise((resolve) => {
+    setupWebsockets(
+      (/* name, uri */) => {
+        // do nothing
```
lencioni: What is this callback for? This comment would be clearer if it explained when we are doing nothing and why we are doing nothing.

src/server/runIOSVisualDiffs.js
```diff
@@ -0,0 +1,108 @@
...
+              height: snapshotImage.height,
+            });
+          });
+      });
+    }
+  });
+}
+
+module.exports = function runIOSVisualDiffs() {
+  const viewportName = 'ios';
```
lencioni: I'm not sure this is quite right. There different sizes on ios, right (e.g. big iphone, small iphone, ipad)? If so, then we'd want this to match at least one of those sizes. Ideally, this would be configurable so you could take screenshots in as many or few of the sizes as you like. Unless maybe that doesn't ever come into play in an ios app? I honestly have no idea how any of that works.

src/server/saveResultToFile.js
```diff
@@ -0,0 +1,18 @@
+function saveResultToFile(runResult) {
```
lencioni: This needs to be exported, eh?

src/server/runIOSVisualDiffs.js
lencioni: Alternatively, maybe `setupWebsockets` should take an object argument instead of positional arguments? That would make this clearer.

src/server/setupWebsockets.js
```diff
@@ -0,0 +1,47 @@
...
+        stories,
+        onStorySnapshot,
+        onFinished
+      );
+    });
+  });
+
+  return new Promise(resolve => {
+    http.listen(5000, function() {
+      console.log('listening on *:5000');
```
lencioni: Should this comment mention what is listening on this port?

native/ios/happo.xcodeproj/xcshareddata/xcschemes/happo.xcscheme
```diff
@@ -0,0 +1,115 @@
...
+      </BuildableProductRunnable>
+   </ProfileAction>
+   <AnalyzeAction
+      buildConfiguration = "Debug">
+   </AnalyzeAction>
+   <ArchiveAction
+      buildConfiguration = "Release"
+      revealArchiveInOrganizer = "YES">
+   </ArchiveAction>
+</Scheme>
```
trotzig: I'm not familiar with iOS development, but I'm assuming that this is generated in some way. Are there any parts of this configuration that "stand out" that are worth mentioning? Or is it mostly a standard setup?

native/ios/happo/AppDelegate.h
```diff
@@ -0,0 +1,16 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
```
trotzig: Do these need to be here?

native/android/app/BUCK
```diff
@@ -0,0 +1,66 @@
...
+# - install Buck
+# - `npm start` - to start the packager
+# - `cd android`
+# - `keytool -genkey -v -keystore keystores/debug.keystore -storepass android -alias androiddebugkey -keypass android -dname "CN=Android Debug,O=Android,C=US"`
+# - `./gradlew :app:copyDownloadableDepsToLibs` - make all Gradle compile dependencies available to Buck
+# - `buck install -r android/app` - compile, install and run application
+#
+
+lib_deps = []
+for jarfile in glob(['libs/*.jar']):
```
trotzig: Do people normally add dependencies to source control or manage them through maven/ivy etc for android apps? This isn't important right now, I'm mostly asking to make sure I'm understanding what's going on.

native/android/gradlew
```diff
@@ -0,0 +1,164 @@
...
+fi
+
+# Split up the JVM_OPTS And GRADLE_OPTS values into an array, following the shell quoting and substitution rules
+function splitJvmOpts() {
+    JVM_OPTS=("$@")
+}
+eval splitJvmOpts $DEFAULT_JVM_OPTS $JAVA_OPTS $GRADLE_OPTS
+JVM_OPTS[${#JVM_OPTS[*]}]="-Dorg.gradle.appname=$APP_BASE_NAME"
+
+exec "$JAVACMD" "${JVM_OPTS[@]}" -classpath "$CLASSPATH" org.gradle.wrapper.GradleWrapperMain "$@"
```
trotzig: Is this copied/generated from somewhere or did you hand-write it? I'm assuming copied/generated.

native/index.js
```diff
@@ -0,0 +1,10 @@
+// this is needed to support socket.io
+window.navigator.userAgent = 'react-native';
+
+const { AppRegistry } = require('react-native');
```
trotzig: Could we move this to the top? Or is the import depending on userAgent to be set?

native/js/HappoJob.js
```diff
@@ -0,0 +1,56 @@
+import React, { Component, PropTypes } from 'react';
+import {
+  View,
+} from 'react-native';
+import io from 'socket.io-client';
+import StoryRenderer from './StoryRenderer';
+import { getStories, getStory } from './StoryManager';
```
trotzig: I like the name "story". Assuming it's the same thing that we use for "example" in the js version, I could consider changing it globally to "story".

native/js/HappoJob.js
```diff
@@ -0,0 +1,56 @@
...
+          />
+        )}
+      </View>
+    );
+  }
+}
+
+HappoJob.propTypes = propTypes;
+
+module.exports = HappoJob;
```
trotzig: Looking at this makes me think that we should switch to using React for rendering examples/stories in the browser version as well.

native/js/Scaffold.js
```diff
@@ -0,0 +1,14 @@
+import React, { Component } from 'react';
+import HappoJob from './HappoJob';
+
+class Scaffold extends Component {
```
trotzig: Stateless function components? Looks like it should work: https://medium.com/front-end-hacking/stateless-components-in-react-native-e9034f2e3701#.z3v1kwmvf

.gitignore
```diff
@@ -3,3 +3,47 @@ lib
...
+#
+build/
+.idea
+.gradle
+local.properties
+*.iml
+
+# node.js
+#
+node_modules/
```
trotzig: lol this list is long 🚋

package.json
```diff
@@ -8,6 +8,8 @@
   "scripts": {
     "build": "npm run --silent clean && babel src/server/ -d lib/server/ --ignore __tests__ && npm run --silent webpack",
     "clean": "rimraf lib public/*.worker.js public/*.bundle.js*",
+    "babel:watch": "babel src/server -d lib/server --ignore __tests__ --watch",
```
trotzig: I've been meaning to add this for a while, thanks! I do think we could extract a separate `babel` step so that we don't have to duplicate the params in `build`.

src/server/runIOSVisualDiffs.js
trotzig: I think I did that to make it easier to test `runVisualDiffs`.

trotzig:
Thanks for working on this! I'm excited to see the progress. 

As someone who hasn't done any iOS or Android development, what are the steps necessary to try this out? You don't have to be specific, just list e.g "install xcode", "pay for developer's license", "read up on objective-c" etc. :)

native/js/HappoJob.js
```diff
@@ -0,0 +1,56 @@
...
+      story: getStory(name),
+    });
+  }
+  onSnapshotCompleted(name, uri) {
+    this.socket.emit('snapshotComplete', { name, uri });
+  }
+  render() {
+    const { story } = this.state;
+    return (
+      <View style={{ paddingTop: 20 }}>
```
lelandrichardson: i didn't check if they did or not, but i was worried that the screenshots would include the status bar

native/js/Scaffold.js
lelandrichardson: they work. i avoid them mostly just because they break hot module reloading

lelandrichardson:
Closing in favor of #185

PR #182, opened 2017-01-12
Prepare for targets

trotzig:
This PR moves most firefox related code into a new folder. It should make the transition into a plugin architecture a little simpler.


trotzig:
Thinking about this some more, I wonder if this should instead be a generic "browser" target? If we add more browsers to the mix, it's probably easiest to just use the firefox setup with a few minor changes. 

Configuration could then look something like this:


```js
const BrowserTarget = require('happo-browser-target');

module.exports = {
  targets: {
    firefox: new BrowserTarget('firefox'),
    safari: new BrowserTarget('safari'),
  }
}
```
With this config, you could run `happo run firefox` or `happo run safari`.

lencioni:
Possibly, but there might be special browsers that don't fit in this plugin, like possibly IE, or mobile safari.

It might be worth keeping them separate and if theres anything shared, that could be a shared dependency of the plugins that need it.

trotzig:
I think most of this got superseded by #183. I'll keep the branch around just in case, but at least this PR won't be merged as-is.

PR #183, opened 2017-01-14
Move to Lerna project structure

lelandrichardson:
to: @lencioni @trotzig 

Overall, this PR breaks the project into 5 packages:

- `happo`: this is essentially the CLI (could also be renamed as `happo-cli` potentially)
- `happo-core`: this is a core module (and has no other happo dependencies) and contains some of the core logic functions for happo like image comparison, and also has the main `config` module.
- `happo-viewer`: this is the website for viewing diffs
- `happo-target-firefox`: this is all of the logic needed to run the current firefox "target".
- `happo-uploader-s3`: this is the uploader plugin that works with s3
- `happo-example-project`: example project

The config structure is now changed so that it requires a `targets` array of "target" objects. A target option has the following shape:

```js
type HappoTarget = {
  name: string,
  run(): Promise<RunResult>,
  debug(): Promise,
};
```

The CLI commands that change as a result of this PR are:

```
happo run [<target>]
```
runs happo with the passed in target name. If target is omitted, all targets run and are merged into a single result

```
happo debug <target>
```
runs `debug` on the passed target name.


package.json
```diff
@@ -41,23 +37,9 @@
...
-    "commander": "^2.9.0",
-    "cssesc": "^1.0.0",
-    "ejs": "^2.5.2",
-    "express": "^4.14.0",
-    "mkdirp": "^0.5.1",
-    "pngjs": "^3.0.0",
-    "ps-node": "^0.1.4",
-    "selenium-webdriver": "^3.0.1",
-    "worker-loader": "^0.7.1"
+    "webpack": "^1.13.2",
```
trotzig: Do we still need webpack here?

packages/happo-core/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "happo-core",
+  "version": "5.0.0",
+  "description": "Command-line tool to visually diff JavaScript components",
```
trotzig: Not sure if this is important, but this might be better as "Core modules used by other happo packages" or something similar.

packages/happo-core/src/getImageFromStream.js
```diff
@@ -0,0 +1,21 @@
+const { PNG } = require('pngjs');
+
+function getImageFromStream(stream) {
+  return new Promise((resolve, reject) => {
+    stream
+      .on('error', reject)
+      .pipe(new PNG())
+      .on('error', reject)
```
trotzig: Unrelated, but I just noticed that `'error'` is listened to twice here.

packages/happo-target-firefox/src/__tests__/checkBrowserVersion-test.js
```diff
@@ -2,7 +2,7 @@ jest.mock('child_process');
 const childProcess = require('child_process');
 
 jest.mock('../config');
-const config = require('../config');
+const { config } = require('happo-core');
```
trotzig: It looks like this fixes a bug as well. I did the same fix in f88d3bb9cb8ef, and unless it's a total pain, maybe we should rebase this PR/branch on top of that as well. 

EDIT - I think rebasing might end up being more of a hassle than it's worth. You should update the `jest.mock` so that the right module is mocked though.

packages/happo-target-firefox/src/index.js
```diff
@@ -0,0 +1,33 @@
...
+module.exports = function firefox(config) {
+  const options = Object.assign({}, DEFAULT_CONFIG, config);
+
+  if (!options.name) {
+    throw new Error('firefox target missing `name` option.');
+  }
+
+  return {
+    name: options.name,
+    debug: () => server.start().then(() => {
```
trotzig: Neat solution to the `debug` cli action, since this is specific to the target.

packages/happo-target-firefox/src/index.js
```diff
@@ -0,0 +1,33 @@
+const { constructUrl } = require('happo-core');
+const runVisualDiffs = require('./runVisualDiffs');
+const initializeWebdriver = require('./initializeWebdriver');
+const closeDriver = require('./closeDriver');
+const checkBrowserVersion = require('./checkBrowserVersion');
+const server = require('./server');
```
trotzig: Maybe the server should be in its own package, or part of core? It's needed for the `review` and `upload` actions, which aren't specific to firefox. Or perhaps we just have two different servers, one for review and one used internally by the firefox target. That's probably a better separation of concerns than trying to cram everything into a single module. 

EDIT: It looks like you already did that. Nice!

packages/happo/src/cli.js
```diff
@@ -0,0 +1,91 @@
...
+
+commander.command('debug <target>').action((target) => {
+  targetForName(target).debug();
+});
+
+commander.command('review-demo').action(() => {
+  // server.start().then(() => {
+  //   console.log(`=> ${constructUrl('/review-demo')}`);
+  // });
+});
```
trotzig: If this turns out to be a pain to maintain, we can throw it out and add it back later on. This has been useful when Joe and I have been iterating on/testing the in-browser diffing implementation, but we're not using it much now.

trotzig:
> The config structure is now changed so that it requires a targets array of "target" objects.

Would it make sense to use an object instead of an array here? I feel like that would be more straightforward. 

```js
module.exports = {
  targets: {
    firefox: firefoxTarget({
      viewports: { ... },
    }),
    reactNativeIOS: reactNativeTarget(),
  }
}
```

Oh, but the `name` is probably useful down the line when generating diff pages for multiple targets? So having it attached to the target itself makes things easier. 

Also since the method creating the target is expected to return an instance, maybe it would be better to have targets as classes? 

```js
const FirefoxTarget = require('happo-target-firefox');

module.exports = {
  targets: [
    new FirefoxTarget({
      viewports: { ... },
    }),
  ],
};
```

lerna.json
```diff
@@ -0,0 +1,7 @@
+{
+  "lerna": "2.0.0-beta.32",
+  "packages": [
+    "packages/*"
+  ],
+  "version": "5.0.0"
```
lencioni: So this is the version of happo, right? Does this automatically bump versions in all of the package.json files for us?

packages/happo-core/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "repository": "https://github.com/Galooshi/happo/tree/master/packages/happo-core",
+  "author": "",
+  "license": "MIT",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "../../node_modules/.bin/babel src -d lib --ignore __tests__",
+    "prepublish": "npm run build"
```
lencioni: Are all of the tests always only run at the top level package?

packages/happo-core/src/getLastResultSummary.js
```diff
@@ -3,10 +3,12 @@ const path = require('path');
...
+function getLastResultSummary() {
   const resultSummaryJSON = fs.readFileSync(
     path.join(config.snapshotsFolder, config.resultSummaryFilename),
     'utf8',
   );
   return JSON.parse(resultSummaryJSON);
-};
+}
+
+module.exports = getLastResultSummary;
```
lencioni: meh, I'd rather we change to `export default` and `import` everywhere but that can certainly wait for another time.

packages/happo-target-firefox/src/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+const DEFAULT_CONFIG = {
+  name: 'firefox',
+};
+
+module.exports = function firefox(config) {
+  const options = Object.assign({}, DEFAULT_CONFIG, config);
```
lencioni: I think we have object spread enabled, FWIW, so this could be

```js
const options = { ...DEFAULT_CONFIG, ...config };
```

packages/happo-target-firefox/src/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+  if (!options.name) {
```
lencioni: This is a fine type of check to have, but it seems a little pointless inside of the target plugin since it is pretty easy to see that the name is there. I think we would get more value out of this check at a higher level--whatever is consuming the target plugins.

package.json
lelandrichardson: from what i gather, typically in a lerna monorepo you put dev deps in the root package.json so they can be shared, and then just specify the deps of each package. webpack here is a dev dep

packages/happo-core/package.json
lelandrichardson: yeah i was copy/pasting package.jsons so the description got carried over for each one. I'll clean up

packages/happo-core/package.json
lelandrichardson: i think so. i'm not really sure. lerna has some commands that let you run commands for all of the packages. `lerna run foo` would run `npm run foo` in each of the packages where a `foo` script is defined. For publishing, lerna lets you publish all of the packages at once. they have an "independent" mode where you can control the versioning of each package independently, or the default which is to have all the packages increment together (which is how babel and react work, for example)

packages/happo-core/src/getLastResultSummary.js
lelandrichardson: that's fine. i wasn't sure since i saw a mixture of `require` / `module.exports` and `import` / `export`

lelandrichardson:
@lencioni @trotzig I moved some things around a bit more. The project now runs and works as far as i can tell. It looks like right now it's consistently marking the 2nd example in the example project as having a diff though. is that expected?

I ended up moving the example project into the packages directory as it makes it easier with all of the automatic symlinking that lerna does for us.

The way I had set up the `config` in `happo-core` was causing a bunch of circular dependencies because the config is a runtime JS file which would probably require a happo library (like `happo-target-firefox`) which would in turn depend on `config` from `happo-core`.

In order to fix the circular dependencies, I made the `config` object just a container for an injectable singleton config.

packages/happo-target-firefox/src/index.js
lelandrichardson: agreed. will move.

packages/happo-target-firefox/src/index.js
lelandrichardson: 👍

lerna.json
lelandrichardson: yes. I was thinking we'd use lerna in `integrated` mode, which increments all versions together (rather than keeping a compatibility matrix in our head)

lelandrichardson:
> Also since the method creating the target is expected to return an instance, maybe it would be better to have targets as classes?

Good call. Updated to reflect this.

lelandrichardson:
> Would it make sense to use an object instead of an array here? I feel like that would be more straightforward.

I think having the target have a "default" name works fine, and then allows users to define more than one of the same target if they like. The expense is an array lookup in the CLI, but i don't think that matters much.

packages/happo-target-firefox/src/HappoRunner.js
```diff
@@ -1,4 +1,4 @@
-import { SCREENSHOT_BOX_ID } from './server/Constants';
+import Constants from 'happo-core/lib/Constants';
```
lelandrichardson: needed to import this from `/lib/Constants` because the module overall has some node-only modules like `fs` and `path` that we don't want to include in the browser.

packages/happo-target-firefox/src/__tests__/checkBrowserVersion-test.js
```diff
@@ -44,18 +36,3 @@ describe('when the version string is unrecognized', () => {
         'Failed to parse Firefox version string "Godzilla Firefox 47.0"');
     }));
 });
-
-describe('when using a different driver', () => {
```
lelandrichardson: since this is now the "firefox target", i removed the logic where the driver could be non-firefox, since i think we will end up implementing that as a separate module

packages/happo-target-firefox/webpack.config.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
   entry: {
-    HappoApp: './src/HappoApp.jsx',
     HappoRunner: './src/HappoRunner.js',
```
lelandrichardson: only the runner is owned by the firefox target. `HappoApp` is in `happo-viewer`

packages/happo/src/cli.js
```diff
@@ -0,0 +1,94 @@
...
+const {
+  config,
+  saveResultsToFile,
+  constructUrl,
+} = require('happo-core');
+const initializeConfig = require('./initializeConfig');
+const validateConfig = require('./validateConfig');
+const { server, uploadLastResult } = require('happo-viewer');
+
+config.set(validateConfig(initializeConfig(process.env.HAPPO_CONFIG_FILE)));
```
lelandrichardson: this is where the config is injected (and validated)

packages/happo/src/validateConfig.js
```diff
@@ -0,0 +1,42 @@
+function fail(message) {
+  /* eslint-disable no-console */
+  console.error('Happo config validation failed:');
+  console.error(message);
+  process.exit(1);
+}
+
+function validateConfig(config) {
```
lelandrichardson: I added some validation code for the config. This will probably help make upgrading easier as well...

packages/happo-target-firefox/src/HappoRunner.js
trotzig: Ah, makes sense.

packages/happo-target-firefox/src/__tests__/checkBrowserVersion-test.js
trotzig: Good call.

trotzig:
> The project now runs and works as far as i can tell. It looks like right now it's consistently marking the 2nd example in the example project as having a diff though. is that expected?

Yep, it has a `Math.random()` in its rendering.

trotzig:
I'm trying this out locally, and so far I haven't run into any issues. 💯 

Just to make sure I understand things correctly, this is what I did to try things out:

1. Checked out your branch
2. Ran `$(npm bin)/lerna bootstrap`
3. Cd:ed into `packages/happo-example-project` and ran `happo run firefox`
4. Watched things work
5. Made some changes in `packages/happo-target-firefox`
6. Re-ran `happo run firefox` in the example project
7. Realized I hadn't started the babel watcher, doh. 
8. Ran `npm run babel -- --watch` from within `happo-target-firefox`
9. Re-ran `happo run firefox` 
10. Saw my changes in action. 

Great work!

packages/happo/src/validateConfig.js
```diff
@@ -0,0 +1,42 @@
...
+  const { targets } = config;
+  if (targets && !Array.isArray(targets)) {
+    fail(`Expected \`targets\` to be an array. Found ${typeof targets} instead.`);
+  }
+
+  const nameSet = Object.create(null);
+
+  targets.forEach((target, index) => {
+    if (!target) {
+      fail(`Targets must be an object. Found ${typeof target} at index ${index} instead.`);
```
trotzig: This should say "Target" instead of "Target".

packages/happo/src/validateConfig.js
trotzig: Yep, good call.

lelandrichardson:
@trotzig yeah, that should do it. I tried to set things up so we didn't need to `cd` into any directories. From the root directory you should be able to:

1. `./example <command>` should run `happo <command>` in the example project
2. `npm run watch` should watch all of the projects at the same time
3. `lerna run build` should build all of the projects at the same time
4. `npm test` should run the tests on the whole project

Perhaps i should make `npm run build` work from the root so that those commands are symmetric

lelandrichardson:
@trotzig @lencioni made `npm` commands for watch and build so people don't necessarily need to know to use lerna now

Will start working on tweaking the readme

packages/happo-target-firefox/src/__tests__/runVisualDiffs-test.js
```diff
@@ -34,37 +57,33 @@ describe('runVisualDiffs', () => {
   });
 
   beforeEach(() => {
-    originalConfig = { ...config };
+    // originalConfig = { ...config };
```
lencioni: should this be deleted?

packages/happo-target-firefox/src/defaultOptions.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
+  name: 'firefox',
+  scriptTimeout: 30000,
+  publicDirectories: [],
+  // TODO: uploader: () => new S3Uploader(),
```
lencioni: TODO

lelandrichardson:
@lencioni @trotzig I added a `happo-uploader-s3` package, and updated some of the READMEs. PTAL!

README.md
```diff
@@ -208,72 +64,60 @@ Happo loads configuration in one of the following ways:
...
 
 Once `happo run` has finished, run `happo review` from the command line. This
 will open a page that compares the latest run's snapshots against the previous
 snapshots.
 
-### `happo debug`
+### `happo debug <targetName>`
 
 If you want to debug rendering your examples, you can run `happo debug`. This
 will open a browser window pointing at `/debug`, listing all your examples. If
```
lencioni: Some of these details should maybe be updated or removed. That can wait until later though.

packages/happo-target-firefox/README.md
```diff
@@ -0,0 +1,239 @@
+# Happo Target: Firefox
+
+## Installation
+
+If you haven't already installed `happo` you will need to do that:
+```bash
+npm install --save-dev happo
+```
+
+To get the firefox target, you will install from npm:
```
lencioni: nit: Firefox

packages/happo-uploader-s3/README.md
```diff
@@ -0,0 +1,66 @@
+# Happo Uploader: Amazon S3
+
+## Installation
+
+If you haven't already installed `happo` you will need to do that:
+```bash
+npm install --save-dev happo
+```
+
+To get the firefox target, you will install from npm:
```
lencioni: This line is incorrect

lelandrichardson:
Note: i'm thinking about adding a gitbook documentation site in a followup PR, as happo will begin to have a larger surface area with more targets/uploaders/ci etc.

lelandrichardson:
@lencioni agreed. I think w/ tests passing and no glaring "we should not do {x}" or "we should fix {y}", we should merge and maybe hold off on publishing for a little bit while we continue to smoke test. What do you think?

lencioni:
@lelandrichardson that sounds good to me.

PR #188, opened 2017-02-01
Process targets serially to prevent collision

lelandrichardson:
to: @lencioni @trotzig 

This is required in the CLI, because when you define multiple targets we want them to run serially rather than in parallel, since they will likely use some of the same ports and what not.

### DIFF
--- a/packages/happo/src/cli.js
+++ b/packages/happo/src/cli.js
@@ -9,6 +9,7 @@ const {
 const initializeConfig = require('./initializeConfig');
 const validateConfig = require('./validateConfig');
 const { server, uploadLastResult } = require('happo-viewer');
+const processSerially = require('./processSerially');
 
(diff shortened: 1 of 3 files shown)


packages/happo/src/cli.js
```diff
@@ -29,12 +29,22 @@ function targetForName(name) {
...
 
+function processSerially(items, fn) {
+  const promises = [];
+  items.reduce((prev, item) => {
+    const promise = prev.then(() => fn(item));
+    promises.push(promise);
+    return promise;
+  }, Promise.resolve());
+  return Promise.all(promises);
+}
```
lencioni: It might not be a bad idea to write a test for this function.

lelandrichardson: seems reasonable. will give it a shot

lelandrichardson:
@lencioni added tests!

PR #194, opened 2017-03-15
Update import-js config for all packages

trotzig:
To make import-js [1] work correctly, we need to configure it a little.

Most of our packages are node projects using CommonJS imports and
exports, so I created a root .importjs.js that package-specific config
then inherited. These projects also transpile code into a `lib` folder,
plus have test files in `__tests__` folders, so it was easy to set up
the excludes configuration (which controls what files import-js won't
import for you).

The happo-viewer package was treated differently since it mostly has
client-side React code where we use es6 imports and exports.

[1]: https://github.com/Galooshi/import-js

### DIFF
--- a/packages/happo-core/.importjs.js
+++ b/packages/happo-core/.importjs.js
@@ -1,0 +1,1 @@
+module.exports = require('../../.importjs.js');
--- a/packages/happo-target-firefox/.importjs.js
+++ b/packages/happo-target-firefox/.importjs.js
@@ -1,0 +1,1 @@
+module.exports = require('../../.importjs.js');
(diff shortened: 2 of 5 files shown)


.importjs.js
```diff
@@ -1,27 +1,8 @@
...
-    return 'import';
-  },
-  namedExports: {
-    'react': ['PropTypes'],
-    'js/src/Diff': [
-      'DiffController',
-    ]
-  },
-  importDevDependencies: true,
+  declarationKeyword: 'const',
```
lencioni: Does this mean we aren't going to use `import` anywhere?

trotzig: I've mostly just mirrored what the code looks like in the project. We can make a sweep to convert to `import`, but I decided that was out of scope for this PR.

trotzig:
Whoops, I cherry-picked this into #198 as well, thinking I never pushed this branch. I'll merge it from there instead.

PR #198, opened 2017-04-04
Fix firefox target

trotzig:
I was setting up happo in a new project and noticed that the latest version wasn't fully working.


trotzig:
@lelandrichardson can you add me (and probably @lencioni as well) as owners to the different npm modules? I'm `trotzig`/`henric.trotzig [at] gmail`. I tried pushing rc.2 but failed because of missing permissions. I kept the commit that was created as part of `lerna publish`, because the `happo` target was successfully pushed. Once I have access, I'll push again.

spikebrehm:
ping @lelandrichardson ^

trotzig:
These commits have been cherry-picked on to master (in https://github.com/Galooshi/happo/commit/f04ea4e).

Issue #178, opened 2017-01-05
Upload images to global location, with filenames that are hashes of file contents


lencioni:
I was chatting with @lelandrichardson about some of our plans for supporting multiple rendering targets (#73) and generating a review HTML page (#176), and we thought it might not be a bad idea to change the way we upload images.

Currently, we upload an HTML file and a set of images as a sort of bundle to a single "directory". Over many runs, this is likely to produce duplicate images. What if, instead, we put the images in a more general location where the filenames are hashes of the file contents. That way, if we have multiple images that are the same, we don't upload them again.

This would have the advantage of allowing us to make all examples browsable on the review page with minimal overhead.

If we do this, the files should be stored in sub-directories so that the directory doesn't end up growing to contain millions of files over time.

We could probably also apply the same technique to the JavaScript files that we need for the review page, to cut down on duplication, but that seems less valuable than doing it for the images.

trotzig:
This is a good idea. Do you have an idea for how it could be implemented as well? It could be as simple as using the `uploader` configuration option that I'm adding/thinking of adding as part of supporting #176. 

```js
// .happo.js
const S3Uploader = require('happo/server/S3Uploader');

module.exports  = {
  uploader: new S3Uploader({ 
    imageDirectory: 'some-global-dir',
  }),
}
```

lencioni:
It would be nice to avoid re-uploading assets that already exist, so we might want to build in that check to the upload process. `imageDirectory` option sounds okay to me, but I think we could probably just set it to something and not make it configurable, right?

trotzig:
Ah, that might be right. Since people mostly have to specify an s3 bucket name. We can choose to make a special folder for images, and check before upload.

lencioni:
We might also try a potential optimization where we upload a JSON file for every sha that maps examples to their images. Then we can potentially download all of the images for the base commit if they already exist. Not sure if that will be faster but it might depending on how long the build takes.

PR #209, opened 2017-06-12
fix(firefox): handle data uris

### LINKED ISSUE #207: waitForImagesToLoad seems to fail on data-uris that contain SVGs
And possibly other types of data-uris. Seeing an error message like this:

```
Failed to load image with url data:image/svg+xml;utf8,<svg><line%20x1=\"100%\"%20y1=\"0\"%20x2=\"0\"%20y2=\"100%\"%20stroke=\"rgba(219,%20219,%20219,%201
```

oliviertassinari:
Closes #207.

Come into that issue trying to add happo to the [storybook](https://github.com/storybooks/storybook) of [react-instantsearch](https://github.com/algolia/react-instantsearch).

### DIFF
--- a/packages/happo-target-firefox/src/__tests__/waitForImagesToRender-test.js
+++ b/packages/happo-target-firefox/src/__tests__/waitForImagesToRender-test.js
@@ -1,0 +1,13 @@
+import { waitForImageToLoad } from '../waitForImagesToRender';
+
+const image = "data:image/svg+xml;utf8,<svg%20viewBox='0%200%2010%2011'%20xmlns='http://www.w3.org/2000/svg'><g%20stroke='%23bfc7d8'%20stroke-width='1.5'%20fill='none'%20fill-rule='evenodd'%20stroke-linecap='round'><path%20d='M5%201.5v8M9%205.5H1'/></g></svg>";
+
+describe('waitForImagesToRender', () => {
+  describe('waitForImageToLoad', () => {
+    it('should handle data uris correctly', () => (
+      waitForImageToLoad(image).then((output) => {
+        expect(output.type).toBe('load');
+      })
+    ));
+  });
+});
--- a/packages/happo-target-firefox/src/findBackgroundImageUrls.js
+++ b/packages/happo-target-firefox/src/findBackgroundImageUrls.js
@@ -5,7 +5,9 @@ export default function findBackgroundImageUrls(string) {
   let match;
   // eslint-disable-next-line no-cond-assign
   while (match = URL_PATTERN.exec(string)) {
-    result.push(match[1]);
+    // We need to unescape the simple quote.
+    const url = match[1].replace(/\\'/g, "'");
+    result.push(url);
   }
   return result;
 }
--- a/packages/happo-target-firefox/src/findBackgroundImageUrls.js
+++ b/packages/happo-target-firefox/src/findBackgroundImageUrls.js
@@ -5,7 +5,8 @@ export default function findBackgroundImageUrls(string) {
   let match;
   // eslint-disable-next-line no-cond-assign
   while (match = URL_PATTERN.exec(string)) {
-    // We need to unescape the simple quote.
+    // Inlined images of the form data:image can contain embedded svg.
+    // To avoid failing to load this image, we need to unescape escaped single quotes.
     const url = match[1].replace(/\\'/g, "'");
     result.push(url);
   }
--- a/packages/happo-target-firefox/src/waitForImagesToRender.js
+++ b/packages/happo-target-firefox/src/waitForImagesToRender.js
@@ -1,6 +1,6 @@
 import findBackgroundImageUrls from './findBackgroundImageUrls';
 
-function waitForImageToLoad(url) {
+export function waitForImageToLoad(url) {
   return new Promise((resolve, reject) => {
     const img = new Image();
     img.onerror = () => reject(new Error(`Failed to load image with url ${url}`));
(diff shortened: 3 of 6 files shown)


packages/happo-target-firefox/src/waitForImagesToRender.js
```diff
@@ -1,7 +1,13 @@
 import findBackgroundImageUrls from './findBackgroundImageUrls';
 
-function waitForImageToLoad(url) {
+export function waitForImageToLoad(url) {
   return new Promise((resolve, reject) => {
+    // No need to wait for the image to be loaded.
+    if (url.indexOf('data:image') === 0) {
```
lencioni: I think we might want to be a little more specific with this search. In my experience this was working alright with JPEG data URIs and was actually needed in order for those to be fully rendered by the browser (although I wonder if we should switch to something like `createImageBitmap` instead). I think this might only be a problem when the data URI is an SVG.

I also wonder if this should be in the promise or if we should add a filter before this function call. Thoughts?

In the project you were trying to add this to, did you encounter this with an SVG data URI or something else?

packages/happo-target-firefox/src/__tests__/waitForImagesToRender-test.js
```diff
@@ -0,0 +1,13 @@
+import { waitForImageToLoad } from '../waitForImagesToRender';
+
+const image = "data:image/svg+xml;utf8,<svg%20viewBox='0%200%2010%2011'%20xmlns='http://www.w3.org/2000/svg'><g%20stroke='%23bfc7d8'%20stroke-width='1.5'%20fill='none'%20fill-rule='evenodd'%20stroke-linecap='round'><path%20d='M5%201.5v8M9%205.5H1'/></g></svg";
```
lencioni: I think you are missing a `>` on the end here.

packages/happo-target-firefox/src/waitForImagesToRender.js
oliviertassinari: I have been digging a bit more in the issue. It seems to be a char encoding issue. I haven't went down the rabbit hole yet. Are you saying that even with not network latency, it's preferable to load the image first?

lencioni: Ah interesting.

>  Are you saying that even with not network latency, it's preferable to load the image first?

Yes, because image rendering is non-instant and non-blocking, so screenshots can actually be taken in the middle of image rendering, which makes it so you end up with a lot of visual diffs for components that use images because the screenshots are taken at different moments while the image was rendering. Here's some more info: https://github.com/Galooshi/happo/issues/162#issuecomment-251465911

oliviertassinari:
I have found a better solution closer to the root of the issue. I have noticed that Firefox escape the `'` (where Chrome don't).

```js
var url1 = window.getComputedStyle(document.querySelector(X)).getPropertyValue('background-image')

// url1 = url("data:image/svg+xml;utf8,<svg%20viewBox=\'0%200%2010%209\'%20xmlns=\'http://www.w3.org/2000/svg\'><path%20d=\'M1%204.88l2.378%202.435L9.046%201.6\'%20stroke-width=\'1.6\'%20stroke=\'%23FFF\'%20fill=\'none\'%20fill-rule=\'evenodd\'%20stroke-linecap=\'round\'%20stroke-linejoin=\'round\'/></svg>")

```

oliviertassinari:
> so let me know if you want any guidance there

Thanks for asking. I'm gonna try to integrate it with Travis CI, however, I'm planning on hijacking his original purpose to only takes the screenshots with Happo and to let [Argos-CI](https://www.argos-ci.com/) handle the diffs and review process.
The only issue I have seen so far with that integration is around the screenshot destination, original test names are base64 making the folder name harder to use. I'm wondering if you would be open adding a function option for generating the screenshots destination.
Finally, I feel like it would be really interesting to build a selenium grid plugin. We have been using [vrtest](https://github.com/nathanmarks/vrtest) on [Material-UI](https://www.argos-ci.com/callemall/material-ui) to take the screenshots in a docker hosting a selenium grid, chrome & firefox, that works pretty well.

oliviertassinari:
Now I have a reproduction example for that issue: https://github.com/algolia/react-instantsearch/pull/130. I have manage to make the 82 passed with the fix of the PR.

packages/happo-target-firefox/src/findBackgroundImageUrls.js
```diff
@@ -5,7 +5,9 @@ export default function findBackgroundImageUrls(string) {
   let match;
   // eslint-disable-next-line no-cond-assign
   while (match = URL_PATTERN.exec(string)) {
-    result.push(match[1]);
+    // We need to unescape the simple quote.
```
trotzig: This comment could be expanded a little to better explain why it's here. How about

> Inlined images of the form `data:image` can contain embedded svg. To avoid failing to load this image, we need to unescape escaped single quotes.

lencioni:
> I'm wondering if you would be open adding a function option for generating the screenshots destination.

I'm hesitant to open this up to be customized because I'm not sure how useful it would be for most people. However, we have been discussing changing how we approach uploading in #178. Do you think something like that would fit your use-case?

oliviertassinari:
Thanks guys :).

> I'm hesitant to open this up to be customized

I can understand, I ended up writing [a normalisation script](https://github.com/argos-ci/argos/blob/master/examples/with-happo/test/normalize.js). Without looking too much in details, I don't think that #178 would answer my issue.

trotzig:
This was released in version `5.0.0-rc.6`. Thanks @oliviertassinari!

PR #214, opened 2017-10-13
[happo-viewer] Bump worker-loader ^0.7.1 -> ^1.0.0

lencioni:
worker-loader pre-1.0.0 has a peerDependency that does not allow webpack
3. This prevents projects that want to update to webpack 3 from getting
onto that version.

The only breaking change listed here is about loader-utils, which I
don't think will impact this project.

https://github.com/webpack-contrib/worker-loader/releases/tag/v1.0.0

### DIFF
--- a/packages/happo-viewer/package.json
+++ b/packages/happo-viewer/package.json
@@ -24,7 +24,7 @@
     "ps-node": "^0.1.4",
     "express": "^4.14.0",
     "ejs": "^2.5.2",
-    "worker-loader": "^0.7.1",
+    "worker-loader": "^1.0.0",
     "cssesc": "^1.0.0"
   }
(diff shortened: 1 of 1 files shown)


kesne:
This LGTM!

lencioni:
It seems that lerna bootstrap is failing in CI. I was able to reproduce locally, but also work around it by going into individual package directories and npm installing there. Now lerna bootstrap works locally. I believe I am running into this issue: https://github.com/lerna/lerna/issues/785

lencioni:
I think this is safe to merge, and I'd like to unblock webpack 3, so I'll be publishing a new release soon. We should possibly follow up with investigating the issues here soon...

PR #215, opened 2017-10-14
Fix rc2 and 3 missing from master

trotzig:
I had accidentally published rc2 and rc3 from a local branch. Sigh. @lencioni noticed when publishing rc.4, and this is my attempt at correcting things. I've cherry-picked commits that were part of the https://github.com/Galooshi/happo/tree/fix-happo-debug branch. 

After merging this, I think it's best to bump versions for all packages (to rc.9).


ljharb:
fwiw, I'd strongly suggest not bumping all versions - i.e., to retain independent versioning. Linked versions creates unnecessary dep churn.

PR #223, opened 2018-06-08
Wait for images defined in srcset attributes

trotzig:
This commit is a port of changes I've made to the new runner for
happo.io. By waiting for images in srcset attributes, we can prevent
spurious diffs.

Instead of trying to figure out which one of the images will be
displayed in the current viewport, I decided to download all. The
tradeoff here being simplicity over performance (dowloading only one of
the images would be faster).

While running tests locally, I noticed that I was getting 2x images
compared to what's in CI. I decided to make the assertions a little
looser to account for both 1x and 2x screenshots.

### DIFF
--- a/packages/happo-target-firefox/src/waitForImagesToRender.js
+++ b/packages/happo-target-firefox/src/waitForImagesToRender.js
@@ -11,10 +11,23 @@ export function waitForImageToLoad(url) {
 
 export default function waitForImagesToRender() {
   return new Promise((resolve, reject) => {
-    const promises = Array.prototype.slice.call(document.querySelectorAll('img'))
+    const images = Array.prototype.slice.call(document.querySelectorAll('img'));
+    const promises = images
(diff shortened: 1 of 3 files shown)


packages/happo-target-firefox/src/waitForImagesToRender.js
```diff
@@ -11,11 +11,24 @@ export function waitForImageToLoad(url) {
...
       .filter(Boolean)
       .map(waitForImageToLoad);
 
+    images.forEach((img) => {
+      const srcset = img.getAttribute('srcset');
+      if (!srcset) {
+        return;
+      }
+
+      srcset.split(/,\s*/).forEach((entry) => {
```
lencioni: I don't think this will work for urls that contain commas

trotzig: Good catch. Fixed in dc21c33.

PR #224, opened 2018-06-10
Fix grabbing urls from srcset attributes

trotzig:
My first attempt at this (312ecd9) was too naive, it didn't take urls
with commas in them into account. In this second attempt, I've tried to
tdd things so that I know we can handle a wider range of values.

I've used https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img
as a guide into figuring out possible values.

### DIFF
--- a/packages/happo-target-firefox/src/getUrlsFromSrcset.js
+++ b/packages/happo-target-firefox/src/getUrlsFromSrcset.js
@@ -1,0 +1,10 @@
+const SRCSET_ITEM = /([^\s]+)(\s+[0-9.]+[wx])?(,?\s*)/g;
+
+export default function getUrlsFromSrcset(value) {
+  const result = [];
+  let match;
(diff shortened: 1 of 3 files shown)


packages/happo-target-firefox/src/getUrlsFromSrcset.js
```diff
@@ -0,0 +1,10 @@
+const SRCSET_ITEM = /([^\s]+)(\s+[0-9.]+[wx])?(,?\s*)/g;
+
+export default function getUrlsFromSrcset(value) {
+  const result = [];
+  let match;
+  while (match = SRCSET_ITEM.exec(value)) { // eslint-disable-line no-cond-assign
```
ljharb: You can use https://npmjs.com/string.prototype.matchall for this :-D

trotzig: Thanks! I'll make that change after I publish.

PR #226, opened 2018-10-24
Wait for images with crossorigin=anonymous

trotzig:
I've had spurious diffs reported from images with a crossorigin
attribute. This could be because the browser kicks off different
requests for the crossorigin image and the preloaded image. This creates
a race condition: if the crossorigin image isn't ready by the time the
preload image is downloaded, it won't show up in the screenshot. We can
fix this by transferring crossorigin information from the original image
to the temporary Image created as part of preloading.

### DIFF
--- a/packages/happo-target-firefox/src/waitForImagesToRender.js
+++ b/packages/happo-target-firefox/src/waitForImagesToRender.js
@@ -1,19 +1,13 @@
 import findBackgroundImageUrls from './findBackgroundImageUrls';
 import getUrlsFromSrcset from './getUrlsFromSrcset';
 
-export function waitForImageToLoad(imageOrUrl) {
+export function waitForImageToLoad({ src, crossOrigin }) {
   return new Promise((resolve, reject) => {
-    let url = imageOrUrl;
-    let crossOrigin;
-    if (typeof imageOrUrl !== 'string') {
-      url = imageOrUrl.src;
-      crossOrigin = imageOrUrl.crossOrigin;
-    }
     const img = new Image();
-    img.onerror = () => reject(new Error(`Happo: Failed to load image with url ${url}`));
+    img.onerror = () => reject(new Error(`Happo: Failed to load image with url ${src}, crossOrigin=${crossOrigin}`));
     img.addEventListener('load', resolve, { once: true });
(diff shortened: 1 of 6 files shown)


packages/happo-target-firefox/src/waitForImagesToRender.js
```diff
@@ -1,11 +1,18 @@
 import findBackgroundImageUrls from './findBackgroundImageUrls';
 import getUrlsFromSrcset from './getUrlsFromSrcset';
 
-export function waitForImageToLoad(url) {
+export function waitForImageToLoad(imageOrUrl) {
```
lencioni: I'm not quite sure I follow why this needs to be polymorphic.

trotzig: It's called from a few more places (urls scraped from `srcset` and `background-image`). 

That being said, I'm not 100% happy with how this looks. Also, it makes me realize I probably need to do this for `srcset` as well.

.travis.yml
```diff
@@ -7,7 +7,7 @@ node_js:
   - "8"
 
 addons:
-  firefox: "56.0"
+  firefox: "62.0"
```
ljharb: Does this mean it’s a breaking change, and we’ll need to update our Firefox version?

trotzig: It shouldn't be based on the changes I made. What version are you on currently? It's possible that all we need is a smaller version bump.

trotzig: At first I thought it might be an update to selenium-webdriver that caused this, but we specify `^3.6.0` and the latest version matching that range is `3.6.0`.

trotzig: Ah, but that leaves `geckodriver`. It has a few versions published after `1.9.0` (which is what we depend on, `^1.9.0`). Let me check if locking that version down affects anything.

ljharb: we’re on 56; we had to update from 52 last time.

trotzig: ah, yeah let's see if we can make things work on 56 then.

trotzig: Okay, locking down geckodriver seems to work. No update required :).

PR #241, opened 2025-10-28
Add freezeAnimations, set default to "last-frame"

trotzig:
This option is best used as last-frame, so we're taking the opportunity to make this a breaking change.

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -263,6 +263,15 @@ interface BaseTarget {
    * with Happo, you may also want to use the `applyPseudoClasses` option.
    */
   allowPointerEvents?: boolean;
+
+  /**
+   * Set `freezeAnimations: 'last-frame'` to freeze the animations at the last
+   * frame. This is the default behavior.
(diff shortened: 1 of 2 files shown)


src/config/index.ts
```diff
@@ -263,6 +263,15 @@ interface BaseTarget {
    * with Happo, you may also want to use the `applyPseudoClasses` option.
    */
   allowPointerEvents?: boolean;
+
+  /**
+   * Set `freezeAnimations: 'last-frame'` to freeze the animations at the last
+   * frame. This is the default behavior.
+   *
+   * Set `freezeAnimations: 'first-frame'` to freeze the animations at the first
+   * frame.
```
lencioni, line 272: Should probably mention the default in the comment

trotzig: It is mentioned just above. 👆🙂

lencioni: Oh! I missed it. I reviewed on my phone.

PR #244, opened 2025-10-29
Update storybook to v10

lencioni:
I did this by running:

```sh
npx storybook@latest upgrade
```

and following the prompts. It installed an eslint plugin, but it seemed
that the types were incompatible so I decided to not worry about that
for now.

Once that was done, there were a few adjustments I needed to make:

- Since storybook v10 only supports ESM, I discovered that we had an
  errant `__dirname` that needed to be updated to `import.meta.dirname`.
- For now I am leaving the minimum storybook version at v9. I'm not
  sure if this will work with storybook v9 or not but hopefully it will
  be fine. If not, we can adjust before we finalize the release.
- The types for the new version of storybook did not include the docgen
  stuff. I think we could add it by following the instructions in
  https://github.com/joshwooding/vite-plugin-react-docgen-typescript but
  I'm not sure why we have that in here so I decided to remove it for
  now.

Closes HAP-483


src/storybook/index.ts
```diff
@@ -28,9 +28,9 @@ function resolveBuildCommandParts() {
 
   const version = getStorybookVersionFromPackageJson();
 
-  if (version < 9) {
+  if (version < 10) {
     throw new Error(
-      `Storybook v${version} is not supported. Please upgrade to v9 or later, or downgrade happo-plugin-storybook to an earlier version.`,
+      `Storybook v${version} is not supported. Please upgrade to storybook v10 or later.`,
```
lencioni: Maybe I'll leave this at 9 for now and see how it goes. We can always tighten it to 10 if we need to before finalizing the release.

lencioni: I left it at 9 and updated the test job to run both v9 and v10

PR #251, opened 2025-11-02
Ensure static package has iframe.html

trotzig:
Since this file plays an important part in a static package, let's make sure it exists.

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -227,16 +227,15 @@ describe('main', () => {
     describe('static integration', () => {
       beforeEach(async () => {
         tmpfs.writeFile(
-          'happo-static/iframe.html',
-          '<!doctype html><html lang="en"><head><meta charset="utf-8" /><script src="/bundle.js"></script></head><body></body></html>',
-        );
-        tmpfs.writeFile(
           'happo.config.ts',
           `
           export default {
             integration: { 
               type: 'static',
-              generateStaticPackage: async () => '${tmpfs.getTempDir()}/happo-static',
+              generateStaticPackage: async () => ({
+                rootDir: '${tmpfs.getTempDir()}/happo-static',
+                entryPoint: 'bundle.js',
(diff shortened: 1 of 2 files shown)


src/cli/__tests__/index.test.ts
```diff
@@ -208,6 +224,54 @@ describe('main', () => {
...
+            apiSecret: 'test-secret',
+          };
+          `,
+        );
+      });
+
+      it('succeeds when iframe.html is found', async () => {
+        await main(['npx', 'happo'], logger);
+        console.log('logger.log.mock.calls', logger.log.mock.calls);
+        console.log('logger.error.mock.calls', logger.error.mock.calls);
```
lencioni: Extra logs

src/network/prepareSnapRequests.ts
```diff
@@ -8,11 +8,31 @@ import deterministicArchive from '../utils/deterministicArchive.ts';
...
+async function createIframeHTML(
+  rootDir: string,
+  entryPoint: string,
+  logger: Logger,
+): Promise<void> {
   const iframePath = path.join(rootDir, 'iframe.html');
-  if (fs.existsSync(iframePath)) {
+
+  if (await fileExists(iframePath)) {
+    logger.info(`Using existing iframe.html at '${iframePath}'`);
```
lencioni, line 32: I added a bit of logging here

src/network/prepareSnapRequests.ts
```diff
@@ -23,15 +43,18 @@ async function createIframeHtml(rootDir: string, entryPoint: string): Promise<vo
     <script src="${entryPoint}"></script>
   </body>
 </html>`;
-  fs.writeFileSync(iframePath, iframeContent);
+
+  await fs.promises.mkdir(rootDir, { recursive: true });
```
lencioni, line 47: Some of the tests were failing on the writeFile because the directory did not exist. I thought we might as well mkdir here just in case.

PR #252, opened 2025-11-03
Make beforeSha a non-null env attribute

trotzig:
I noticed a job I was running tried to post something against /api/reports/undefined/compare/foobar. To make this less likely to happen, I'm defaulting the beforeSha to afterSha. This will reduce the likelyhood of us creating comparisons etc with a faulty beforeSha. To prevent comparisons against equal shas, I'm adding a conditional in createAsyncComparison.

### DIFF
--- a/src/network/createAsyncComparison.ts
+++ b/src/network/createAsyncComparison.ts
@@ -47,6 +47,12 @@ export default async function createAsyncComparison(
   }: EnvironmentResult,
   logger: Logger,
 ): Promise<CreateAsyncComparisonResult> {
+  if (beforeSha === afterSha) {
+    throw new Error(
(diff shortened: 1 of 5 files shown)


src/network/createAsyncComparison.ts
```diff
@@ -46,7 +46,11 @@ export default async function createAsyncComparison(
     fallbackShas,
   }: EnvironmentResult,
   logger: Logger,
-): Promise<CreateAsyncComparisonResult> {
+): Promise<CreateAsyncComparisonResult | null> {
+  if (beforeSha === afterSha) {
+    return null;
```
lencioni: I am wondering if we should throw here instead, so we don't end up hiding a env resolution bug on accident.

Also, if beforeSha is undefined, this won't be right either.

trotzig: I made this throw.

PR #253, opened 2025-11-03
Create iframe.html automatically for static integrations

trotzig:
So that consumers don't have to worry about this, we'll let them return an object with the path to the output dir plus the name of the entrypoint.

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -57,12 +57,14 @@ interface StaticIntegration {
   type: 'static';
 
   /**
-   * An async function that generates a static package. Returns the path to the
-   * folder containing the static files. You can use a prebuilt package (simply
-   * point to the output directory of the Storybook build) or build the package
-   * yourself.
+   * An async function that generates a static package. Returns an object with
+   * the path to the folder containing the static files and the path to the
+   * entry point file relative to the root directory.
+   *
+   * @example
+   * { rootDir: 'dist/static', entryPoint: 'index.js' }
    */
-  generateStaticPackage: () => Promise<string>;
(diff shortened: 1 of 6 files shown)


src/config/index.ts
```diff
@@ -57,12 +57,14 @@ interface StaticIntegration {
...
-   * An async function that generates a static package. Returns the path to the
-   * folder containing the static files. You can use a prebuilt package (simply
-   * point to the output directory of the Storybook build) or build the package
-   * yourself.
+   * An async function that generates a static package. Returns an object with
+   * the path to the folder containing the static files and the path to the
+   * entry point file relative to the root directory.
+   *
+   * @example
+   * { rootDir: 'dist/static', entryPoint: 'index.js' }
```
lencioni, line 65: We should flesh out this example a bit more soon, so it shows that you might need to execute a build tool on some directory and then return path to the files that were built.

src/network/prepareSnapRequests.ts
```diff
@@ -1,15 +1,38 @@
...
 import generateStorybookStaticPackage from '../storybook/index.ts';
 import deterministicArchive from '../utils/deterministicArchive.ts';
 import Logger, { logTag } from '../utils/Logger.ts';
 import uploadAssets from './uploadAssets.ts';
 
+async function createIframeHtml(rootDir: string, entryPoint: string): Promise<void> {
+  const iframePath = path.join(rootDir, 'iframe.html');
+  if (fs.existsSync(iframePath)) {
+    return;
+  }
```
lencioni, line 15: We need to make sure to document that you can override the `iframe.html` file by generating it in the build directory, and why you might want to (e.g. different lang, different meta viewport, etc)

PR #258, opened 2025-11-05
Rename static integration to custom integration

lencioni:
Static is a confusing name, because this is sort of the opposite of static--we are building a JavaScript bundle and shipping that off to the workers to be executed. This is anything but static. After brainstorming a few different names, we thought custom worked the best for this.


src/config/index.ts
```diff
@@ -53,18 +53,18 @@ interface PlaywrightIntegration {
...
+   * An async function that generates a custom package. Returns an object with
+   * the path to the folder containing the custom files and the path to the
    * entry point file relative to the root directory.
    *
    * @example
-   * { rootDir: 'dist/static', entryPoint: 'index.js' }
+   * { rootDir: 'dist/custom', entryPoint: 'index.js' }
    */
-  generateStaticPackage: () => Promise<{ rootDir: string; entryPoint: string }>;
+  build: () => Promise<{ rootDir: string; entryPoint: string }>;
```
lencioni, line 67: I thought `build` was a better name for this function.

src/config/RemoteBrowserTarget.ts
```diff
@@ -37,8 +37,13 @@ export interface CSSBlock {
...
 interface ExecuteParams {
   globalCSS?: string | Array<CSSBlock>;
-  assetsPackage?: unknown;
-  staticPackage?: unknown;
+
+  /** Path to the assets package */
+  assetsPackage?: string;
+
+  /** Path to the static package */
+  staticPackage?: string;
```
lencioni, line 45: This is what the worker is expecting, so I decided to keep it named staticPackage here.

src/config/index.ts
trotzig: Agreed!

PR #261, opened 2025-11-05
Add waitForContent to pages

lencioni:
I was working on updating documentation and noticed we didn't include this option.


lencioni:
@trotzig I realized that we aren't testing the pages integration yet. We need to add that!

PR #263, opened 2025-11-06
Add pages test integration

trotzig:
While working on this I realized that `waitForSelector` wasn't listed in the options, so I added that.


package.json
```diff
@@ -125,6 +126,7 @@
     "eslint-plugin-depend": "^1.3.1",
     "eslint-plugin-simple-import-sort": "^12.1.1",
     "eslint-plugin-unicorn": "^62.0.0",
+    "http-server": "^14.1.1",
```
lencioni: Is this used?

trotzig: Good catch -- no! I first added it but realized I can't use local pages...

trotzig: I pushed a fix.

PR #265, opened 2025-11-06
Replace HAPPO_BASE_BRANCH with --baseBranch CLI arg

lencioni:
We have decided to use CLI args instead of env vars wherever possible.
Since this env var is only consumed by the CLI, it makes a great
candidate.

To do this, we need to pass the CLI args down to resolve environment so
the value can be used at the correct moment.

The default value in the help text is slightly misleading--it will be
origin/main unless being used in Azure, in which case it will try to
default to `SYSTEM_PULLREQUEST_TARGETBRANCH` and fall back to
origin/main. I'm not sure of a good way to make that clear in the help
text at this time though.

To help ensure that this works well, we need the resolveEnvironment
function to be consistent. We were calling it in the e2e controller to
support running Cypress and Playwright without the Happo wrapper, but
since that does not have the CLI args passed to it you could end up in a
confusing situation where the resolveEnvironment gives you different
results. We decided that this whole situation was unnecessary and
complex and are just removing that whole code path instead. This means
that if you want Happo screenshots with Cypress or Playwright, you must
run it via the happo command.

Part of HAP-504

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -45,6 +45,10 @@ function parseRawArgs(rawArgs: Array<string>) {
         type: 'string',
         short: 'c',
       },
+
+      baseBranch: {
+        type: 'string',
+      },
     },
 
     allowPositionals: true,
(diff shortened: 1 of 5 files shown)


lencioni:
We decided to remove that resolveEnvironment call instead. This PR has been updated!

PR #279, opened 2025-11-07
Replace HAPPO_DOWNLOAD_ALL with config option

trotzig:
Instead of having this magical env variable, we instead allow people to configure their cypress/playwright integrations with a boolean downloadAllAssets.

While looking at tests, I decided to ditch the custom server and use startTestServer instead. But I had to make the port static since we assert some hashes in the package. With a dynamic port, it was hard to know what the hashes would be.

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -41,7 +41,7 @@ interface BaseE2EIntegration {
 
   /**
    * Whether to download and include all assets in the asset package. If false
-   * (default), only local assets will be included. WHen true, even external
+   * (default), only local assets will be included. When true, even external
    * assets will be included.
    */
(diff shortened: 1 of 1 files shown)


src/config/index.ts
```diff
@@ -33,22 +33,26 @@ export interface StorybookIntegration {
...
   allowFailures?: boolean;
-}
-
-interface PlaywrightIntegration {
-  type: 'playwright';
 
   /**
-   * Whether to allow failures.
+   * Whether to download and include all assets in the asset package. If false
+   * (default), only local assets will be included. WHen true, even external
```
lencioni:
```suggestion
   * (default), only local assets will be included. When true, even external
```

PR #281, opened 2025-11-07
Replace archiver and adm-zip with fflate

lencioni:
I found this package that small, fast, and does the thing we need. Looks like a win to me.


lencioni:
I think this may have broken images in playwright and cypress. See here: https://happo.io/a/8/p/2081/compare/22c3f756bd8599186b41873e469a52aacef9f2b1/5d71ae54e0238eac00656f8bb56532add23f31e2?jobId=1950550

lencioni:
Fix: https://github.com/happo/happo/pull/286

PR #285, opened 2025-11-07
Refactor Logger to use node's built-in styleText

lencioni:
Documentation here:

https://nodejs.org/docs/latest-v24.x/api/util.html#utilstyletextformat-text-options

This detects whether color is supported or not already, so we can drop the supports-color dependency. Nice!


src/utils/__tests__/Logger.test.ts
```diff
@@ -0,0 +1,215 @@
+import assert from 'node:assert';
```
lencioni, line 1: I copied the test over from happo.io and asked cursor to rewrite it for node test runner.

PR #297, opened 2025-11-14
Improve package.json metadata for storybook integration catalog

lencioni:
Documentation on this can be found here:

  https://storybook.js.org/docs/addons/integration-catalog#addon-metadata

For supported frameworks, while we haven't tested this with everything in the list, I'm not aware of anything that would prevent it from working with any of these frameworks, so I decided to just use the entire list.

One requirement listed for addons, is this:

> preset.js file written as an ES5 module at the root level

While we have a preset.js file, it is not at the root level of the package. I am hoping that it will be okay though. We'll see!

Part of HAP-526

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "happo",
   "version": "6.0.1",
-  "description": "Visual regression testing and accessibility testing",
+  "description": "Catch unexpected visual and accessibility changes and UI bugs",
   "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -145,16 +145,35 @@
...
+      "react",
+      "react-native",
+      "svelte",
+      "vue",
+      "web-components"
+    ],
+    "unsupportedFrameworks": []
+  },
   "keywords": [
+    "storybook-addon",
```
lencioni, line 165: They say this needs to be the very first keyword for some reason, so I put it here instead of keeping the alphabetical order.

package.json
```diff
@@ -145,16 +145,35 @@
     "mime-types": "^3.0.1",
     "srcset": "^5.0.2"
   },
+  "storybook": {
+    "displayName": "Happo",
+    "icon": "https://happo.io/static/happo-hippo.png",
```
lencioni, line 150: On closer inspection, I'm not sure this is a good image for this. It is not very crisp and it seems to have some background texture of some sort.

lencioni: Oh maybe the texture is added by my browser on transparent backgrounds.

PR #315, opened 2025-11-28
Use temporary API key/secret locally

trotzig:
If there is no API key or secret configured, we can ask the user for one. I've worked on a change in the happo server that will show a page with an "Authenticate" button. This button will cause a short-lived API key to be created. This key will then be communicated back to the CLI script using a localhost server and a redirect.

### DIFF
--- a/src/config/getShortLivedAPIToken.ts
+++ b/src/config/getShortLivedAPIToken.ts
@@ -113,6 +113,7 @@ export default async function getShortLivedAPIToken(
 
   try {
     // Open browser
+    console.log(`Opening URL: ${authUrl}`);
     await openBrowser(authUrl);
     const result = await callbackPromise;
     return result;
--- a/src/config/loadConfig.ts
+++ b/src/config/loadConfig.ts
@@ -92,7 +92,6 @@ async function getFallbackApiToken(
       logger.log(
         `Failed to obtain temporary pull-request token for URL: ${environment.link}`,
       );
-      return undefined;
     }
   }
 


src/config/getShortLivedAPIToken.ts
```diff
@@ -0,0 +1,155 @@
...
+        
+      </style>
+    </head>
+    <body>
+      <main>
+        <h1>${title}</h1> 
+        <p>${body}</p>
+      </main>
+    </body>
+  </html>`;
```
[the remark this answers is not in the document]
trotzig: We're in full control of this page and I don't see a future where we are not.

src/config/getShortLivedAPIToken.ts
```diff
@@ -0,0 +1,155 @@
...
+  } catch (error) {
+    logger.error(
+      `Failed to authenticate: ${error instanceof Error ? error.message : String(error)}`,
+    );
+    return null;
+  } finally {
+    // Clean up server
+    await serverInfo.close();
+  }
+}
```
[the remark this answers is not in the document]
trotzig: I added a test.

src/config/promptUser.ts
```diff
@@ -0,0 +1,21 @@
+import { createInterface } from 'node:readline';
+
+export default function promptUser(message: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const rl = createInterface({
+      input: process.stdin,
+      output: process.stdout,
```
lencioni, line 7: If we allow input and output to be passed in, we should be able to use real streams in our tests and avoid the mocking. This will make the tests better.

src/config/__tests__/getShortLivedAPIToken.test.ts
```diff
@@ -0,0 +1,197 @@
...
+let mockOpenBrowser: MockOpenBrowser;
+mock.module('../openBrowser.ts', {
+  defaultExport: (() => {
+    mockOpenBrowser = mock.fn(async (url: string) => {
+      // Store the URL that was passed
+      mockOpenBrowser.lastUrl = url;
+    }) as MockOpenBrowser;
+    return mockOpenBrowser;
+  })(),
+});
```
lencioni, line 45: I'd like to avoid mocking entirely or mock at the lowest level possible. I posted some comments later about how we might achieve that.

src/config/openBrowser.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (platform === 'darwin') {
+    command = `open "${url}"`;
+  } else if (platform === 'win32') {
+    command = `start "" "${url}"`;
+  } else {
+    // Linux and others
+    command = `xdg-open "${url}"`;
+  }
+
+  return execAsync(command).then(() => {
```
lencioni, line 19: I found this similar implementation (https://github.com/azhar22k/ourl/blob/main/index.js) that does some things a little differently that I think we should consider:

1. It uses `execFile` instead of `exec` which should be a little more efficient and secure.
2. It encodes the URL, which seems like it might not be a bad idea.
3. It uses `cmd /c start` for windows. I'm not sure if this is important or not, but it seems reasonable.

Alternatively, I think we should consider bringing in `open` as a dependency instead of rolling this ourselves: https://github.com/sindresorhus/open

Also, it would be nice to avoid mocking this entire module and instead mock the exec--possibly via dependency injection by passing in an exec function as an argument with a default value of node's exec function.

src/config/getShortLivedAPIToken.ts
```diff
@@ -0,0 +1,155 @@
...
+  const serverInfo = await startServer((req, res) => {
+    // Get port from the request socket
+    const url = new URL(req.url ?? '', `http://localhost:${serverInfo.port}`);
+
+    if (url.pathname === '/callback') {
+      const key = url.searchParams.get('key');
+      const secret = url.searchParams.get('secret');
+
+      if (key && secret) {
+        // Send success response
```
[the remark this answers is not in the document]
lencioni: This seems like a reasonable thing to do

src/config/loadConfig.ts
```diff
@@ -70,9 +71,41 @@ async function getPullRequestSecret(
...
+      );
+      return {
+        key: environment.link,
+        secret: pullRequestSecret,
+      };
+    } catch {
+      logger.log(
+        `Failed to obtain temporary pull-request token for URL: ${environment.link}`,
+      );
+      return undefined;
```
lencioni: What do you think about removing this return? That will allow it to use both fallbacks if possible.

src/config/getShortLivedAPIToken.ts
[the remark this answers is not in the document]
trotzig: Done

src/config/openBrowser.ts
trotzig: Let's push this to a follow-up

src/config/loadConfig.ts
trotzig: Yeah that seems like the right move.

Issue #340, opened 2025-12-20
v6 does not post a GitHub comment on a standard `happo` CLI run


sslogar-plaid:
In the process of upgrading from `happo-ci` v13 to `happo` v6, and if I'm tracing code correctly, there seems to be a breaking change where a GitHub comment will not be posted on a non-E2E test run.

`happo-ci`:
1. Runs [`compare`](https://github.com/happo/happo.io/blob/062355fd8ca9ea42288453ff04d5d4b0da4c02c7/bin/happo-ci#L152) by default
2. `compare` always [posts to GH](https://github.com/happo/happo.io/blob/062355fd8ca9ea42288453ff04d5d4b0da4c02c7/src/executeCli.js#L142-L147)

`happo`:
1. Runs [`handleDefaultCommand`](https://github.com/happo/happo/blob/76934ab971c526652995fc5ab6a66a5b22cc2c95/src/cli/index.ts#L232-L235) by default
2. Command [creates a comparison and exits](https://github.com/happo/happo/blob/76934ab971c526652995fc5ab6a66a5b22cc2c95/src/cli/index.ts#L282-L289)
3. Other commands, like `happo -- playwright test` and `happo finalize` have [logic](https://github.com/happo/happo/blob/76934ab971c526652995fc5ab6a66a5b22cc2c95/src/e2e/wrapper.ts#L101-L113) [branches](https://github.com/happo/happo/blob/76934ab971c526652995fc5ab6a66a5b22cc2c95/src/e2e/wrapper.ts#L169-L185) which post a GH comment, but these can't be used outside of E2E

Steps to repro:
1. `pnpm install happo@6.2.0`
2. `pnpm happo --link='https://github.com/foo/bar/pull/123' --githubToken='123'`
3. Report is outputted, but no comment is posted, nor do logs indicate an attempt was made:
```
[HAPPO] Async report URL: https://happo.io/XXX
[HAPPO] Async comparison URL: https://happo.io/XXX
```

PR #347, opened 2026-01-07
Post GitHub comment for default command

trotzig:
When conditions are met, we want to make sure that the GitHub comment is posted. Before, his was only happening in the e2e case.

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -59,6 +59,16 @@ mock.module('../../network/uploadAssets.ts', {
   },
 });
 
+const postGitHubCommentMock: Mock<
+  typeof import('../../network/postGitHubComment.ts').default
+> = mock.fn(async () => {
+  return true;
+});
+
+mock.module('../../network/postGitHubComment.ts', {
+  defaultExport: postGitHubCommentMock,
+});
+
 // Install fresh mocks & imports for each test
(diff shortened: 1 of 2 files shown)


trotzig:
Fixes #340

src/cli/__tests__/index.test.ts
```diff
@@ -211,6 +222,118 @@ describe('main', () => {
       );
     });
 
+    it('posts GitHub comment when conditions are met', async () => {
```
lencioni, line 225: It would be best to be more explicit than "when conditions are met" here

src/cli/index.ts
```diff
@@ -284,6 +284,21 @@ async function handleDefaultCommand(
...
       );
       logger.log(`[HAPPO] Async comparison URL: ${asyncComparison.compareUrl}`);
+
+      if (environment.link && environment.githubToken && config.githubApiUrl) {
+        // githubToken and githubApiUrl are set which means that we should post
+        // a comment to the PR.
+        // https://docs.happo.io/docs/continuous-integration#posting-statuses-without-installing-the-happo-github-app
+        const postGitHubComment = (await import('../network/postGitHubComment.ts'))
+          .default;
+        await postGitHubComment({
```
lencioni, line 294: I was wondering if we want to log something, but it looks like the postGitHubComment function does log stuff so 👍🏻

PR #354, opened 2026-01-14
Add deepCompare configuration support

trotzig:
Add support for configuring deep compare settings in the Happo config file. The deepCompare option allows users to customize image comparison behavior with the following settings:

- compareThreshold (required): Threshold for comparing images (0-1)
- diffAlgorithm (optional): Algorithm to use ("color-delta" or "ssim", defaults to "color-delta")
- ignoreThreshold (optional): Threshold for ignoring pixel differences (0-1)
- ignoreWhitespace (optional): Whether to ignore whitespace in diffs
- applyBlur (optional): Whether to apply blur to smooth differences

The deepCompare object is only sent in compare requests when configured, allowing remote defaults to apply when not specified.

This change depends on a change to the Happo API (currently in progress).

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -108,6 +108,50 @@ interface PagesIntegration {
   pages: Array<Page>;
 }
 
+/**
+ * Settings for deep compare functionality
+ */
+export interface DeepCompareSettings {
+  /**
+   * Threshold for comparing images with the given diff algorithm (float between
+   * 0 and 1). 1 means all differences are allowed. 0 means no differences are
+   * allowed. A good starting value is 0.03 for color-delta and 0.01 for ssim.
+   */
+  compareThreshold: number;
+
+  /**
+   * Algorithm to use for diff comparison. Must be "color-delta" or "ssim".
+   * Defaults to "color-delta" if not provided. Note that "ssim" is experimental
+   * and may be removed in the future.
+   */
+  diffAlgorithm?: 'color-delta' | 'ssim';
+
+  /**
+   * Threshold for ignoring individual pixel differences, side-stepping the
+   * compare threshold. Used relatively to the image size. E.g. a value of 0.01
+   * means 1% of the pixels can be above the compare threshold. Use this option
+   * if your screenshots contain images or graphics with sharp noise. It is not
+   * recommended to use this option for other types of diffs. (float
+   * between 0 and 1).
+   */
+  ignoreThreshold?: number;
+
+  /**
+   * Whether to ignore whitespace in the diff. If true, whitespace differences
+   * will not be considered when comparing images. Whitespace is defined as a
+   * vertical section in a screenshot containing a single solid color.
+   */
+  ignoreWhitespace?: boolean;
+
+  /**
+   * Whether to apply blur to the diff. This can be used to smooth out subtle
+   * differences that would otherwise be above the compare threshold. This
+   * should mainly be used when your screenshots have a high contrast and you
+   * want to smooth out some of the sharpness that can otherwise cause flakiness.
+   */
+  applyBlur?: boolean;
+}
+
 export interface Config {
   /**
    * Key used to authenticate with the Happo API. Never store this in plain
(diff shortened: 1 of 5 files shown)


src/config/index.ts
```diff
@@ -108,6 +108,41 @@ interface PagesIntegration {
   pages: Array<Page>;
 }
 
+/**
+ * Settings for deep compare functionality
+ */
+export interface DeepCompareSettings {
+  /**
+   * Threshold for comparing images with the given diff algorithm (float between 0 and 1).
```
lencioni: It might be good to provide some guidance here around what is a normal range or good default for this value.

Also make it clear that 1 will ignore every diff and 0 will ignore no diffs.

src/config/index.ts
```diff
@@ -108,6 +108,41 @@ interface PagesIntegration {
...
+   */
+  compareThreshold: number;
+
+  /**
+   * Algorithm to use for diff comparison. Must be "color-delta" or "ssim".
+   * Defaults to "color-delta" if not provided.
```
lencioni: Should probably mention that ssim is considered to be experimental and maybe link to documentation.

src/config/index.ts
```diff
@@ -108,6 +108,41 @@ interface PagesIntegration {
...
+   */
+  diffAlgorithm?: 'color-delta' | 'ssim';
+
+  /**
+   * Threshold for ignoring individual pixel differences, side-stepping the compare threshold.
+   * Used relatively to the image size. E.g. a value of 0.01 means 1% of the pixels can be above
+   * the compare threshold. (float between 0 and 1).
+   */
+  ignoreThreshold?: number;
```
lencioni, line 137: I find this whole business to be confusing

src/config/index.ts
```diff
@@ -108,6 +108,41 @@ interface PagesIntegration {
...
+
+  /**
+   * Whether to ignore whitespace in the diff. If true, whitespace differences will not be
+   * ignored when comparing images.
```
lencioni: It might be worth mentioning how whitespace is defined here.

src/config/index.ts
```diff
@@ -108,6 +108,41 @@ interface PagesIntegration {
...
+   */
+  ignoreWhitespace?: boolean;
+
+  /**
+   * Whether to apply blur to the diff. This can be used to smooth out subtle differences
+   * that would otherwise be above the compare threshold.
```
lencioni: It would be good to provide some guidance here about when you might want to do this. Probably true for all of these options.

src/config/index.ts
```diff
@@ -163,6 +198,11 @@ export interface Config {
     | PlaywrightIntegration
     | CustomIntegration
     | PagesIntegration;
+
+  /**
+   * An object with settings for deep compare.
```
lencioni, line 212: It would be good to briefly explain what deep compare means.

src/config/loadConfig.ts
```diff
@@ -103,6 +103,83 @@ async function getFallbackApiToken(
...
+    return;
+  }
+
+  if (typeof deepCompare !== 'object' || Array.isArray(deepCompare)) {
+    throw new TypeError(
+      `Invalid \`deepCompare\` in config file ${configFilePath}: must be an object, got: ${Array.isArray(deepCompare) ? 'array' : typeof deepCompare}.`,
+    );
+  }
+
+  const settings = deepCompare as Record<string, unknown>;
```
lencioni: Can we narrow the type by adding more checks instead of using type casting?

src/config/loadConfig.ts
```diff
@@ -103,6 +103,83 @@ async function getFallbackApiToken(
   return undefined;
 }
 
+function validateDeepCompareSettings(
+  deepCompare: unknown,
+  configFilePath: string,
+): void {
```
lencioni: It might be good to change the return type here to an `asserts` type, so this function will narrow the type of `deepCompare`.

src/config/index.ts
trotzig: Yeah agreed. It's used quite a bit across Happo projects but it might not be useful for many of them. It was added to help a customer reduce flake coming from a canvas rendering thing. It would have 1-3 sharp pixel differences in the screenshots and I couldn't come up with a better solution at the time. I'll update the description and let you re-review to see if it clears up? We could consider removing support for it but that's a bigger task (we'd have to make sure that we're not creating more flake).

src/config/index.ts
trotzig: I added more info for each option.

PR #360, opened 2026-01-16
Make assertElement work across realms

### LINKED ISSUE #358: happoScreenshot() cannot be chained to valid Element
**Context**: upgrading to new consolidated package, `6.2.3`, specifically a Cypress integration. 
________

**Issue**: 

instances of `HappoScreenshot()` which previously worked started to fail with the following error: 

Image

[originating in happo/src/browser/assertElement.ts,line 28]

The specific check/assertion that is failing is this: 

```
  const getCustomApp= () => cy.get('[id="single-spa-application:custom-app"]', { timeout: 20000 });
   getCustomApp().happoScreenshot({
       // some options here
    });
```

Interestingly, when I used `HappoScreenshot()` with other Elements, no error was thrown. Example: `cy.get('.custom-icon').happoScreenshot();`

________

**Potential cause/solution:** 

It seems that because Cypress can have multiple windows/iframes the `instanceof gt.Element` check will fail if the Element has been created in a different window. (Because it checks if it's been created by or according to that specific Element class, not just that it's a generic Element). 

A simple repro (per Copilot) to show that  `instanceof Element` can evaluate as true or false depending on which window `Element` is pulled from: 

```
it('repro: element created in iframe has different Element constructor', () => {
  cy.visit('/');

  cy.window().then(win => {
    const iframe = win.document.createElement('iframe');
    win.document.body.appendChild(iframe);

    const iframeWin = iframe.contentWindow!;
    const div = iframeWin.document.createElement('div');
    // @ts-ignore
    expect(div instanceof iframeWin.Element).to.eq(true);
    // @ts-ignore
    expect(div instanceof window.Element).to.eq(false);
  });
});
```

If this is the root cause, then I think a check of `nodeType === 1` would avoid this and still provide the type safety
________

**Misc**
I found this article/documentation describing some pitfalls of `instanceof` in Cypress, in case it's useful: https://glebbahmutov.com/cypress-examples/recipes/instanceof.html

Let me know if you need me to clarify anything or if I'm missing something, of course.

lencioni:
The assertElement function can fail incorrectly in Cypress, due to Cypress's use of different window and document objects. More details here:

https://glebbahmutov.com/cypress-examples/recipes/instanceof.html

This means that instanceof checks should probably be avoided in Cypress.

We can still preserve type safety by being just a little more careful with how we go about things here.

Since we are already checking the nodeType, which is just a constant 1, we don't really need the instanceof check at all for Element.

We can do something similar for our NodeList check, but since all we really care about is whether we can iterate over it, we can do a slightly different check for that behavior specifically.

I was unable to come up with a good test that would fail the instanceof NodeList check, but I think that should be okay.

Fixes #358

### DIFF
--- a/src/browser/__tests__/assertElement.test.ts
+++ b/src/browser/__tests__/assertElement.test.ts
@@ -15,42 +15,42 @@ describe('assertElement', () => {
     initDOM('<!DOCTYPE html>');
     const { document: doc } = globalThis.window;
     const el = doc.createTextNode('text node');
-    assert.throws(() => assertElement(el, globalThis.window));
+    assert.throws(() => assertElement(el));
   });
 
   it('does not throw if the element is an HTMLElement', () => {
     initDOM('<!DOCTYPE html>');
     const { document: doc } = globalThis.window;
     const el = doc.createElement('div');
-    assert.doesNotThrow(() => assertElement(el, globalThis.window));
+    assert.doesNotThrow(() => assertElement(el));
   });
 
   it('does not throw if the element is an MathMLElement', () => {
     initDOM('<!DOCTYPE html>');
     const { document: doc } = globalThis.window;
     const el = doc.createElement('math');
-    assert.doesNotThrow(() => assertElement(el, globalThis.window));
+    assert.doesNotThrow(() => assertElement(el));
   });
 
   it('does not throw if the element is an SVGElement', () => {
     initDOM('<!DOCTYPE html>');
     const { document: doc } = globalThis.window;
     const el = doc.createElementNS('http://www.w3.org/2000/svg', 'svg');
-    assert.doesNotThrow(() => assertElement(el, globalThis.window));
+    assert.doesNotThrow(() => assertElement(el));
   });
 
   it('does not throw if the element is an array of elements', () => {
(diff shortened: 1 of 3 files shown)


src/browser/__tests__/assertElement.test.ts
```diff
@@ -15,42 +15,42 @@ describe('assertElement', () => {
...
-    assert.doesNotThrow(() => assertElement(el, globalThis.window));
+    assert.doesNotThrow(() => assertElement(el));
   });
 
   it('does not throw if the element is an array of elements', () => {
     initDOM('<!DOCTYPE html>');
     const { document: doc } = globalThis.window;
     const el = doc.createElement('div');
-    assert.doesNotThrow(() => assertElement([el], globalThis.window));
+    assert.doesNotThrow(() => assertElement([el]));
```
trotzig, line 46: Does this need to be an array?

trotzig (approved):

src/browser/__tests__/assertElement.test.ts
lencioni: Yeah it's testing the array case

PR #373, opened 2026-01-27
Add `flake` CLI command

trotzig:
This command will hand off to the Happo API and the new Get flakes endpoint (https://happo.io/docs/api#getFlakes). This command can be useful if you are using an LLM to help with addressing flake.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -26,6 +26,36 @@ function parseDashdashCommandParts(
   return rawArgs.slice(dashdashIndex + 1);
 }
 
+const fallbackOptions = {
+  allProjects: {
+    type: 'boolean',
+  },
+  format: {
+    type: 'string',
+  },
+  project: {
+    type: 'string',
+  },
+  limit: {
+    type: 'string',
+  },
+  page: {
+    type: 'string',
+  },
+  component: {
+    type: 'string',
+  },
+  variant: {
+    type: 'string',
+  },
+  target: {
+    type: 'string',
+  },
+  sha: {
+    type: 'string',
+  },
+} as const;
+
 function parseRawArgs(rawArgs: Array<string>) {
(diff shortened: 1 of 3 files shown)


src/cli/index.ts
```diff
@@ -89,6 +89,42 @@ function parseRawArgs(rawArgs: Array<string>) {
       githubToken: {
         type: 'string',
       },
+
+      allProjects: {
```
lencioni: This is a lot of new options. It might be nice to organize the options in the code a bit to make it easier to understand right here what it is there for? e.g. add a variable like `flakeOptions` for these and spread that in to the list. Maybe same for other options? Just a bit of variable naming, ya know?

src/cli/index.ts
```diff
@@ -124,6 +161,17 @@ Options:
   --nonce <nonce>       Nonce to use for Cypress/Playwright comparison
   --githubToken <token> GitHub token to use for posting Happo statuses as comments. Use in combination with the \`githubApiUrl\` configuration option. (default: auto-detected from environment)
 
+Flake command options:
+  --allProjects        List flakes across all projects (default: current project)
```
lencioni:
```suggestion
  --allProjects         List flakes across all projects (default: current project)
```

src/cli/index.ts
```diff
@@ -124,6 +161,17 @@ Options:
...
+  --format <format>     Output format for flake command (default: human, use "json" for raw output)
```
lencioni:
```suggestion
  --format <format>     Output format for flake command (default: "human", use "json" for raw output)
```

lencioni (approved):

src/cli/index.ts
trotzig: Good idea!

PR #458, opened 2026-04-10
Add --skip CLI argument for partial runs

trotzig:
## Summary

- Adds a \`--skip\` CLI argument accepting a JSON array of \`{component, variant?}\` objects to skip in a run
- When skipped examples are provided, happo calls \`POST /api/reports/:sha/find-baseline\` to find a baseline SHA; if found, creates an extends-report snap request (\`POST /api/snap-requests/extends-report\`) with \`{extendedSnaps, extendsSha, project}\`; if no baseline is found, falls back to a full run
- Injects a \`<script type="application/json" id="happo-skipped">\` data block into \`iframe.html\` for Storybook/Custom integrations; writes a temp file and passes \`HAPPO_SKIP_FILE\` env var for Cypress/Playwright (avoids env var size limits)
- Shared \`parseSkip\` / \`validateSkip\` / \`toSkipSet\` / \`isInSkipSet\` helpers in \`src/isomorphic/parseSkip.ts\`; O(1) skip lookups via \`SkipSet\`
- Omitting \`variant\` in a skip item skips all variants for that component
- Removes the legacy \`skip\` config option from \`StorybookIntegration\` in favour of the CLI argument
- Adds \`scripts/getSkip.ts\` and \`pnpm test:storybook:skipped\` for daily-rotating partial test runs

## Test plan

- [ ] \`pnpm build:types\` passes with no errors
- [ ] \`pnpm test\` passes (325 tests, 0 failures)
- [ ] \`pnpm test:storybook:skipped\` runs a partial storybook run with one example skipped (requires API credentials)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -40,7 +40,7 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           happo-api-key: ${{ secrets.HAPPO_API_KEY }}
           happo-api-secret: ${{ secrets.HAPPO_API_SECRET }}
-          projects: 'cypress,storybook-v9,storybook-v10,playwright,playwright-nonce,custom,pages'
+          projects: 'cypress,storybook-v9,storybook-v10,storybook-skipped,playwright,playwright-nonce,custom,pages'
 
   cypress:
     runs-on: ubuntu-latest
@@ -86,6 +86,14 @@ jobs:
       - run: pnpm test:storybook
         env:
           HAPPO_STORYBOOK_VERSION: ${{ matrix.version }}
+
+  storybook-skipped:
+    runs-on: ubuntu-latest
+    name: Storybook (skipped examples)
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-env
+      - run: pnpm test:storybook:skipped
 
   playwright:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -83,7 +83,7 @@ jobs:
       - uses: ./.github/actions/setup-env
       - if: matrix.dependencies != ''
         run: pnpm add -D ${{ matrix.dependencies }}
-      - run: pnpm test:storybook
+      - run: pnpm test:storybook -- --skippedExamples "$(node ./scripts/getSkippedExamples.ts)"
         env:
           HAPPO_STORYBOOK_VERSION: ${{ matrix.version }}
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -83,7 +83,7 @@ jobs:
       - uses: ./.github/actions/setup-env
       - if: matrix.dependencies != ''
         run: pnpm add -D ${{ matrix.dependencies }}
-      - run: pnpm test:storybook -- --skippedExamples "$(node ./scripts/getSkippedExamples.ts)"
+      - run: pnpm build:dist && node dist/cli/main.js -c ./happoconfigs/happo.storybook.config.ts --skippedExamples "$(node ./scripts/getSkippedExamples.ts)"
         env:
           HAPPO_STORYBOOK_VERSION: ${{ matrix.version }}
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -83,7 +83,7 @@ jobs:
       - uses: ./.github/actions/setup-env
       - if: matrix.dependencies != ''
         run: pnpm add -D ${{ matrix.dependencies }}
-      - run: pnpm build:dist && node dist/cli/main.js -c ./happoconfigs/happo.storybook.config.ts --skippedExamples "$(node ./scripts/getSkippedExamples.ts)"
+      - run: pnpm test:storybook
         env:
           HAPPO_STORYBOOK_VERSION: ${{ matrix.version }}
 
--- a/scripts/getSkippedExamples.ts
+++ b/scripts/getSkippedExamples.ts
@@ -1,0 +1,23 @@
+/**
+ * Outputs a --partial JSON argument for a storybook partial run, cycling
+ * through a set of examples based on the current day of the week so that a
+ * different example is skipped on each day.
+ *
+ * Usage:
+ *   node scripts/getPartialExamples.ts
+ */
+
+const examples = [
+  { component: 'Stories', variant: 'ButtonWithText' },
+  { component: 'Stories', variant: 'MiscLarge' },
+  { component: 'Stories', variant: 'ButtonFirefoxOnly' },
+  { component: 'Stories', variant: 'Portal' },
+  { component: 'Stories', variant: 'ButtonWithImage' },
+  { component: 'Stories', variant: 'ButtonWithSomeEmoji' },
+  { component: 'Stories', variant: 'Lazy' },
+];
+
+const day = new Date().getDay(); // 0 (Sun) – 6 (Sat)
+const skipped = examples[day % examples.length];
+
+process.stdout.write(JSON.stringify([skipped]));
--- a/scripts/getSkippedExamples.ts
+++ b/scripts/getSkippedExamples.ts
@@ -8,12 +8,12 @@
  */
 
 const examples = [
-  { component: 'Stories', variant: 'ButtonWithText [white]' },
-  { component: 'Stories', variant: 'MiscLarge [white]' },
-  { component: 'Stories', variant: 'ButtonFirefoxOnly [white]' },
+  { component: 'Stories', variant: 'Button With Text [white]' },
+  { component: 'Stories', variant: 'Misc Large [white]' },
+  { component: 'Stories', variant: 'Button Firefox Only [white]' },
   { component: 'Stories', variant: 'Portal [white]' },
-  { component: 'Stories', variant: 'ButtonWithImage [white]' },
-  { component: 'Stories', variant: 'ButtonWithSomeEmoji [white]' },
+  { component: 'Stories', variant: 'Button With Image [white]' },
+  { component: 'Stories', variant: 'Button With Some Emoji [white]' },
   { component: 'Stories', variant: 'Lazy [white]' },
 ];
 
--- a/scripts/getSkippedExamples.ts
+++ b/scripts/getSkippedExamples.ts
@@ -1,10 +1,10 @@
 /**
- * Outputs a --partial JSON argument for a storybook partial run, cycling
- * through a set of examples based on the current day of the week so that a
- * different example is skipped on each day.
+ * Outputs a --skippedExamples JSON argument, cycling through a set of
+ * examples based on the current day of the week so that a different example
+ * is skipped on each day.
  *
  * Usage:
- *   node scripts/getPartialExamples.ts
+ *   node scripts/getSkippedExamples.ts
  */
 
 const examples = [
--- a/scripts/getSkippedExamples.ts
+++ b/scripts/getSkippedExamples.ts
@@ -8,13 +8,13 @@
  */
 
 const examples = [
-  { component: 'Stories', variant: 'ButtonWithText' },
-  { component: 'Stories', variant: 'MiscLarge' },
-  { component: 'Stories', variant: 'ButtonFirefoxOnly' },
-  { component: 'Stories', variant: 'Portal' },
-  { component: 'Stories', variant: 'ButtonWithImage' },
-  { component: 'Stories', variant: 'ButtonWithSomeEmoji' },
-  { component: 'Stories', variant: 'Lazy' },
+  { component: 'Stories', variant: 'ButtonWithText [white]' },
+  { component: 'Stories', variant: 'MiscLarge [white]' },
+  { component: 'Stories', variant: 'ButtonFirefoxOnly [white]' },
+  { component: 'Stories', variant: 'Portal [white]' },
+  { component: 'Stories', variant: 'ButtonWithImage [white]' },
+  { component: 'Stories', variant: 'ButtonWithSomeEmoji [white]' },
+  { component: 'Stories', variant: 'Lazy [white]' },
 ];
 
 const day = new Date().getDay(); // 0 (Sun) – 6 (Sat)
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -5,6 +5,7 @@ import type { ConfigWithDefaults } from '../config/index.ts';
 import { findConfigFile, loadConfigFile } from '../config/loadConfig.ts';
 import type { EnvironmentResult } from '../environment/index.ts';
 import resolveEnvironment from '../environment/index.ts';
+import { validateSkippedExamples } from '../isomorphic/parseSkippedExamples.ts';
 import type { Logger, SkipItem } from '../isomorphic/types.ts';
 import type { ParsedCLIArgs } from './parseOptions.ts';
 import { parseOptions } from './parseOptions.ts';
@@ -244,29 +245,12 @@ export async function main(
     if (args.dashdashCommandParts) {
       let validatedSkippedExamplesJSON: string | undefined;
       if (environment.skippedExamples) {
-        let parsed: unknown;
         try {
-          parsed = JSON.parse(environment.skippedExamples);
+          validateSkippedExamples(environment.skippedExamples);
         } catch (e) {
-          logger.error('[HAPPO] Failed to parse --skippedExamples JSON:', e);
-          process.exitCode = 1;
-          return;
-        }
-        if (
-          !Array.isArray(parsed) ||
-          !parsed.every(
-            (item) => {
-              if (typeof item !== 'object' || item === null) return false;
-              const r = item as Record<string, unknown>;
-              return (
-                typeof r.component === 'string' &&
-                (r.variant === undefined || typeof r.variant === 'string')
-              );
-            },
-          )
-        ) {
           logger.error(
-            '[HAPPO] --skippedExamples must be a JSON array of {component, variant?} objects',
+            '[HAPPO] Invalid --skippedExamples:',
+            e instanceof Error ? e.message : String(e),
           );
           process.exitCode = 1;
           return;
@@ -373,34 +357,16 @@ async function handleDefaultCommand(
         return;
       }
 
-      let parsed: unknown;
       try {
-        parsed = JSON.parse(environment.skippedExamples);
+        skippedExamples = validateSkippedExamples(environment.skippedExamples);
       } catch (e) {
-        logger.error('[HAPPO] Failed to parse --skippedExamples JSON:', e);
-        process.exitCode = 1;
-        return;
-      }
-
-      if (
-        !Array.isArray(parsed) ||
-        !parsed.every((item): item is SkipItem => {
-          if (typeof item !== 'object' || item === null) return false;
-          const r = item as Record<string, unknown>;
-          return (
-            typeof r.component === 'string' &&
-            (r.variant === undefined || typeof r.variant === 'string')
-          );
-        })
-      ) {
         logger.error(
-          '[HAPPO] --skippedExamples must be a JSON array of {component, variant?} objects',
+          '[HAPPO] Invalid --skippedExamples:',
+          e instanceof Error ? e.message : String(e),
         );
         process.exitCode = 1;
         return;
       }
-
-      skippedExamples = parsed;
 
       const findBaselineReport = (
         await import('../network/findBaselineReport.ts')
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -255,15 +255,18 @@ export async function main(
         if (
           !Array.isArray(parsed) ||
           !parsed.every(
-            (item) =>
-              typeof item === 'object' &&
-              item !== null &&
-              typeof (item as Record<string, unknown>).component === 'string' &&
-              typeof (item as Record<string, unknown>).variant === 'string',
+            (item) => {
+              if (typeof item !== 'object' || item === null) return false;
+              const r = item as Record<string, unknown>;
+              return (
+                typeof r.component === 'string' &&
+                (r.variant === undefined || typeof r.variant === 'string')
+              );
+            },
           )
         ) {
           logger.error(
-            '[HAPPO] --skippedExamples must be a JSON array of {component, variant} objects',
+            '[HAPPO] --skippedExamples must be a JSON array of {component, variant?} objects',
           );
           process.exitCode = 1;
           return;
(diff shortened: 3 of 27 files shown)


.github/workflows/test.yaml
```diff
@@ -83,10 +83,18 @@ jobs:
       - uses: ./.github/actions/setup-env
       - if: matrix.dependencies != ''
         run: pnpm add -D ${{ matrix.dependencies }}
-      - run: pnpm test:storybook
+      - run: pnpm build:dist && node dist/cli/main.js -c ./happoconfigs/happo.storybook.config.ts --skippedExamples "$(node ./scripts/getSkippedExamples.ts)"
```
lencioni: Was this intentional? Do we want to use `--skippedExamples` here in the regular (non-skipped) storybook jobs?

src/cli/index.ts
```diff
@@ -127,6 +127,7 @@ Options:
   --notify <emails>     One or more (comma-separated) email addresses to notify with results
   --nonce <nonce>       Nonce to use for Cypress/Playwright comparison
   --githubToken <token> GitHub token to use for posting Happo statuses as comments. Use in combination with the \`githubApiUrl\` configuration option. (default: auto-detected from environment)
+  --skippedExamples <json> JSON array of {component, variant} objects to skip in this run and borrow from the nearest baseline report instead
```
lencioni: Is it possible to pass just component without variant and skip all variants for that component? That seems like it would be the most likely use case, and being specific about the variants seems less useful and less likely.

Also, maybe we should shorten this to `--skip`?

I'm also thinking about scenarios where you maybe have a lot of things to list, like 10,000. We might need to provide a way for that to be passed in e.g. via a file path that contains the data in it. We can come back to that if we need to later.

src/custom/index.ts
```diff
@@ -42,26 +42,41 @@ const happoStatic = {
...
-        const example = examples[currentIndex];
-
-        if (!example) {
-          // we're done
-          return;
-        }
-
-        if (example.render) {
-          await example.render();
+        const happoSkipped = (globalThis as { happoSkipped?: Array<{ component: string; variant: string }> }).happoSkipped;
```
lencioni: It would be nice to see if we can avoid typecasting

src/custom/index.ts
```diff
@@ -42,26 +42,41 @@ const happoStatic = {
...
+          const example = examples[currentIndex];
+
+          if (!example) {
+            // we're done
+            return;
+          }
+
+          if (
+            Array.isArray(happoSkipped) &&
+            happoSkipped.some(
```
lencioni: Maybe we should convert happoSkipped to a Set or something with constant lookup time?

src/cypress/index.ts
```diff
@@ -201,16 +202,23 @@ Cypress.Commands.add(
...
+        cachedSkippedExamples = happoSettings?.skippedExamples ?? [];
+        if (cachedSkippedExamples.some((item) => item.component === component && item.variant === variant)) {
+          return;
+        }
         takeSnapshot(cachedAutoApplyPseudoStateAttributes);
       });
     } else {
+      if (cachedSkippedExamples?.some((item) => item.component === component && item.variant === variant)) {
+        return;
+      }
```
lencioni: Same here, maybe cachedSkippedExamples should be a Set?

src/cypress/task.ts
```diff
@@ -139,13 +140,34 @@ const task: HappoTask = {
     return null;
   },
 
-  happoGetIntegrationConfig(): IntegrationConfig {
+  happoGetIntegrationConfig(): HappoScreenshotConfig {
     const integration = controller.config?.integration;
+    let skippedExamples: Array<{ component: string; variant: string }> = [];
+    if (process.env.HAPPO_SKIPPED_EXAMPLES) {
+      try {
+        const parsed: unknown = JSON.parse(process.env.HAPPO_SKIPPED_EXAMPLES);
```
lencioni: I worry about this env var being too big. Maybe we should use a tmp file instead?

src/cypress/task.ts
```diff
@@ -139,13 +140,34 @@ const task: HappoTask = {
...
+        if (
```
lencioni: maybe we should log something if the form of `parsed` is not what we expect

src/network/createExtendsReportSnapRequest.ts
```diff
@@ -0,0 +1,29 @@
...
+        extendsSha,
+        project: config.project,
+      },
+    },
+    config,
+    { retryCount: 3 },
+  );
+
+  if (!result || !('requestId' in result) || typeof result.requestId !== 'number') {
+    throw new Error('Invalid response from extends-report snap request API');
```
lencioni: Do we want to include anything from the response in this error?

src/network/findBaselineReport.ts
```diff
@@ -0,0 +1,36 @@
...
+        },
+      },
+      config,
+      { retryCount: 2 },
+      logger,
+    );
+
+    if (result && 'sha' in result && typeof result.sha === 'string') {
+      return result.sha;
+    }
```
lencioni, line 30: Do we want to throw if the result is not what we expect?

src/network/prepareSnapRequests.ts
```diff
@@ -56,14 +56,35 @@ interface BuildPackageResult {
   estimatedSnapsCount?: number;
 }
 
+async function injectSkippedIntoIframe(
+  iframePath: string,
+  skipped: Array<{ component: string; variant: string }>,
+): Promise<void> {
+  const content = await fs.promises.readFile(iframePath, 'utf8');
+  const skippedScript = `<script type="text/javascript">window.happoSkipped = ${JSON.stringify(skipped)};</script>`;
```
lencioni: Why not type="application/json"? Just need to escape `</script>` with that approach.

src/playwright/index.ts
```diff
@@ -103,6 +103,28 @@ export const test: TestType<
         ? (integration.autoApplyPseudoStateAttributes ?? false)
         : false;
 
+    const partialSkipped: Array<{ component: string; variant: string }> = (() => {
+      if (!process.env.HAPPO_SKIPPED_EXAMPLES) return [];
+      try {
+        const parsed: unknown = JSON.parse(process.env.HAPPO_SKIPPED_EXAMPLES);
+        if (
+          Array.isArray(parsed) &&
+          parsed.every(
```
lencioni: This seems duplicated from earlier in this PR. Maybe we should have a shared helper?

src/playwright/index.ts
```diff
@@ -128,6 +150,10 @@ export const test: TestType<
         throw new Error('Missing `variant`');
       }
 
+      if (partialSkipped.some((item) => item.component === component && item.variant === variant)) {
```
lencioni: Maybe use a Set here too?

.github/workflows/test.yaml
trotzig: This is here temporarily (there's a "temp:" commit that I will revert).

src/cli/index.ts
trotzig: Yeah. My first attempt had the flag named `--partial` and you could specify different ways to make it partial. That might be the way to go: 

- `--partial '{ "only": [{"file": "src/.stories/Button.stories.js"}] }'` in case user wants to use inclusion in storybook (similar to `--onlyStoryFiles`). 
- `--partial '{ "skip": [{ "component": "Button"] }'` in case user wants to use exclusion but only specify the component
- `--partial '{ "skip": [{ "component": "Button", "variant": "default" ] }'` in case user wants to use exclusion for a single component+variant example.

src/network/findBaselineReport.ts
trotzig: We want to at least log the result. We still need to proceed (using the fallback: whole report) when the response is unexpected.

src/network/prepareSnapRequests.ts
trotzig: Will do.

src/playwright/index.ts
trotzig: Will do.

src/custom/index.ts
trotzig: I'll try.

src/custom/index.ts
trotzig: Will try.

src/cypress/index.ts
trotzig: Will attempt.

src/network/createExtendsReportSnapRequest.ts
trotzig: Will do.

src/cli/index.ts
lencioni: meh, I think `--skip` and `--only` would be nicer than `--partial`, but I just kinda dislike encoding JSON into CLI args.

src/cli/index.ts
```diff
@@ -255,15 +255,18 @@ export async function main(
...
           !Array.isArray(parsed) ||
           !parsed.every(
-            (item) =>
-              typeof item === 'object' &&
-              item !== null &&
-              typeof (item as Record<string, unknown>).component === 'string' &&
-              typeof (item as Record<string, unknown>).variant === 'string',
+            (item) => {
+              if (typeof item !== 'object' || item === null) return false;
+              const r = item as Record<string, unknown>;
```
lencioni: Can we avoid typecasting here?

src/cli/index.ts
```diff
@@ -381,16 +384,17 @@ async function handleDefaultCommand(
...
-        !parsed.every(
-          (item): item is SkipItem =>
-            typeof item === 'object' &&
-            item !== null &&
-            typeof (item as Record<string, unknown>).component === 'string' &&
-            typeof (item as Record<string, unknown>).variant === 'string',
-        )
+        !parsed.every((item): item is SkipItem => {
+          if (typeof item !== 'object' || item === null) return false;
+          const r = item as Record<string, unknown>;
```
lencioni: Here too (this seems to be duplicated code)

src/cypress/task.ts
```diff
@@ -142,18 +142,19 @@ const task: HappoTask = {
...
           parsed.every(
-            (item): item is { component: string; variant: string } =>
+            (item): item is { component: string; variant?: string } =>
               typeof item === 'object' &&
               item !== null &&
-              typeof item.component === 'string' &&
-              typeof item.variant === 'string',
+              typeof (item as Record<string, unknown>).component === 'string' &&
+              ((item as Record<string, unknown>).variant === undefined ||
+                typeof (item as Record<string, unknown>).variant === 'string'),
```
lencioni: And the typecasting here too

src/playwright/index.ts
```diff
@@ -103,18 +103,19 @@ export const test: TestType<
         ? (integration.autoApplyPseudoStateAttributes ?? false)
         : false;
 
-    const partialSkipped: Array<{ component: string; variant: string }> = (() => {
+    const partialSkipped: Array<{ component: string; variant?: string }> = (() => {
```
lencioni: Maybe import SkipItem?

src/playwright/index.ts
```diff
@@ -103,18 +103,19 @@ export const test: TestType<
...
           parsed.every(
-            (item): item is { component: string; variant: string } =>
+            (item): item is { component: string; variant?: string } =>
               typeof item === 'object' &&
               item !== null &&
-              typeof item.component === 'string' &&
-              typeof item.variant === 'string',
+              typeof (item as Record<string, unknown>).component === 'string' &&
+              ((item as Record<string, unknown>).variant === undefined ||
+                typeof (item as Record<string, unknown>).variant === 'string'),
```
lencioni: More typecasting and duplication

src/storybook/index.ts
```diff
@@ -96,7 +96,7 @@ export default async function buildStorybookPackage({
   usePrebuiltPackage = false,
   skippedExamples,
 }: Omit<StorybookIntegration, 'type'> & {
-  skippedExamples?: Array<{ component: string; variant: string }>;
+  skippedExamples?: Array<{ component: string; variant?: string }>;
```
lencioni: SkipItem type import?

src/cli/index.ts
```diff
@@ -321,13 +327,54 @@ async function handleDefaultCommand(
...
+    let baselineSha: string | undefined;
+
+    if (environment.skippedExamples) {
+      try {
+        skippedExamples = JSON.parse(environment.skippedExamples) as Array<SkipItem>;
+      } catch (e) {
+        logger.error('[HAPPO] Failed to parse --skippedExamples JSON:', e);
+        process.exitCode = 1;
+        return;
+      }
```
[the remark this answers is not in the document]
trotzig: Fixed — added shape validation (must be a JSON array of `{component, variant?}` objects) in both the default and e2e CLI paths. Invalid input now exits with a clear error message before any API calls are made.

src/cli/index.ts
```diff
@@ -321,13 +327,54 @@ async function handleDefaultCommand(
...
+      ).default;
+      baselineSha = await findBaselineReport(environment, config, logger);
+      if (!baselineSha) {
+        logger.log(
+          '[HAPPO] No baseline report found for --skippedExamples run. Generating a full report instead.',
+        );
+        skippedExamples = undefined;
+      }
+    }
+
```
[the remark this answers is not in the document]
trotzig: Fixed — `--skippedExamples` now explicitly rejects unsupported integration types (only `storybook` and `custom` are supported in the default command path) and exits with an error message listing the supported types.

src/cli/index.ts
```diff
@@ -243,6 +248,7 @@ export async function main(
         args.dashdashCommandParts,
         configFilePath,
         logger,
+        environment.skippedExamples,
       );
```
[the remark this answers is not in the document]
trotzig: Fixed — the e2e path now validates the JSON and shape of `--skippedExamples` before spawning the child process, exiting with an error on malformed input.

src/playwright/index.ts
```diff
@@ -103,6 +103,18 @@ export const test: TestType<
         ? (integration.autoApplyPseudoStateAttributes ?? false)
         : false;
 
+    const partialSkipped: Array<{ component: string; variant: string }> = (() => {
+      if (!process.env.HAPPO_SKIPPED_EXAMPLES) return [];
+      try {
+        return JSON.parse(process.env.HAPPO_SKIPPED_EXAMPLES) as Array<{
+          component: string;
+          variant: string;
+        }>;
```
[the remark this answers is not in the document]
trotzig: Fixed — replaced with a call to the new shared `parseSkippedExamples` helper in `src/isomorphic/parseSkippedExamples.ts`, which includes `Array.isArray` and per-item shape validation.

src/cypress/task.ts
```diff
@@ -139,13 +140,22 @@ const task: HappoTask = {
     return null;
   },
 
-  happoGetIntegrationConfig(): IntegrationConfig {
+  happoGetIntegrationConfig(): HappoScreenshotConfig {
     const integration = controller.config?.integration;
+    let skippedExamples: Array<{ component: string; variant: string }> = [];
+    if (process.env.HAPPO_SKIPPED_EXAMPLES) {
+      try {
+        skippedExamples = JSON.parse(process.env.HAPPO_SKIPPED_EXAMPLES) as typeof skippedExamples;
```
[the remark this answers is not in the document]
trotzig: Fixed — same as above; now uses the shared `parseSkippedExamples` helper.

src/network/prepareSnapRequests.ts
```diff
@@ -56,14 +56,32 @@ interface BuildPackageResult {
...
 
+async function injectSkippedIntoIframe(
+  iframePath: string,
+  skipped: Array<{ component: string; variant: string }>,
+): Promise<void> {
+  const content = await fs.promises.readFile(iframePath, 'utf8');
+  const injected = content.replace(
+    '<head>',
+    `<head><script type="text/javascript">window.happoSkipped = ${JSON.stringify(skipped)};</script>`,
+  );
```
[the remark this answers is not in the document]
trotzig: Fixed — now uses a case-insensitive `/<head\b[^>]*>/i` regex and throws with a descriptive error if no match is found.

scripts/getSkippedExamples.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Outputs a --partial JSON argument for a storybook partial run, cycling
+ * through a set of examples based on the current day of the week so that a
+ * different example is skipped on each day.
+ *
+ * Usage:
+ *   node scripts/getPartialExamples.ts
```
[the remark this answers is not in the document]
trotzig: Fixed — comment updated to reference `--skippedExamples` and `getSkippedExamples.ts`.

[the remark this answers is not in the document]
trotzig: Fixed.

src/custom/index.ts
trotzig: Done. In `src/custom/index.ts`, the skipped examples are now read from a `<script type="application/json" id="happo-skipped">` element injected into `iframe.html`, using `document.getElementById('happo-skipped')?.textContent` — no typecast needed. The code is guarded with `typeof document === 'undefined' ? null : ...` for non-browser environments.

src/custom/index.ts
trotzig: Done. Now using a `SkipSet = readonly [Set<string>, Set<string>]` — one set for component-only items, one for `"component\0variant"` keys. The `isInSkipSet` helper checks both in O(1). This is shared via `src/isomorphic/parseSkippedExamples.ts`.

src/cypress/index.ts
trotzig: Done. `cachedSkippedExamples` is now `cachedSkipSet: SkipSet | null` (the same `readonly [Set<string>, Set<string>]` type). Lookup uses `isInSkipSet(cachedSkipSet, component, variant)`.

src/cypress/task.ts
trotzig: Done. In `happoGetIntegrationConfig` in `src/cypress/task.ts`, we now warn via `console.warn` if `HAPPO_SKIPPED_EXAMPLES` is not valid JSON or if it parses to a non-array value.

src/network/createExtendsReportSnapRequest.ts
trotzig: Done. The error message now includes `JSON.stringify(result)`: `Invalid response from extends-report snap request API: ${JSON.stringify(result)}`.

src/network/findBaselineReport.ts
trotzig: Done. `findBaselineReport` now throws on an unexpected response shape. The throw is caught by the surrounding try/catch in `prepareSnapRequests.ts`, which logs the error and returns `undefined` — falling back to using the full report.

src/network/prepareSnapRequests.ts
trotzig: Good call. Changed to `type="application/json"` with the `</script>` → `<\/script>` escape via `String.raw` + `replaceAll`. No JavaScript execution required, just a data block.

src/playwright/index.ts
trotzig: Done. Created `src/isomorphic/parseSkippedExamples.ts` with `parseSkippedExamples`, `toSkipSet`, and `isInSkipSet`. Now shared by Playwright, Cypress, and the custom integration.

src/playwright/index.ts
trotzig: Done. Playwright now uses `toSkipSet(parseSkippedExamples(process.env.HAPPO_SKIPPED_EXAMPLES))` and `isInSkipSet(skipSet, component, variant)` from the shared `src/isomorphic/parseSkippedExamples.ts` module.

src/cypress/task.ts
trotzig: Yep, will do.

trotzig: Done. The wrapper now writes the JSON to a temp file (`happo-skipped-${process.pid}.json` in `os.tmpdir()`) and passes `HAPPO_SKIPPED_EXAMPLES_FILE` pointing to it. Cypress and Playwright read the file with `fs.readFileSync`. The temp file is deleted in the `finally` block.

src/cli/index.ts
trotzig: Done. Both validation paths in cli/index.ts now use validateSkippedExamples() from the shared module — no typecasts.

src/cli/index.ts
trotzig: Done. Extracted validateSkippedExamples (throws on invalid input) into src/isomorphic/parseSkippedExamples.ts; both validation paths in cli/index.ts now call it instead of duplicating the logic.

src/cypress/task.ts
trotzig: Done. cypress/task.ts now reads the file written by the wrapper and passes the raw string directly to parseSkippedExamples — no inline parsing or typecasts.

src/playwright/index.ts
trotzig: Done. Removed the partialSkipped variable entirely; now calling toSkipSet(parseSkippedExamples(rawSkipped)) with the SkipSet type from the shared module.

src/playwright/index.ts
trotzig: Done. All inline parsing and typecasting is gone; reading from HAPPO_SKIPPED_EXAMPLES_FILE and delegating entirely to the shared parseSkippedExamples helper.

src/storybook/index.ts
trotzig: Done. Imported SkipItem from isomorphic/types.ts and used it in the function signature.

src/cli/index.ts
```diff
@@ -321,13 +342,66 @@ async function handleDefaultCommand(
   await startJob(config, environment, logger);
 
   try {
+    // When --skippedExamples is set, first resolve the baseline SHA. If no
+    // baseline is found we fall back to a full run (no skipping).
+    let skippedExamples: Array<SkipItem> | undefined;
+    let baselineSha: string | undefined;
+
+    if (environment.skippedExamples) {
```
lencioni: I wonder if this code of validating the skipped examples belongs in the environment code instead of here in the CLI?

src/isomorphic/parseSkip.ts
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * A pair of Sets used for O(1) skip lookups.
+ * - [0]: component-only skips (match all variants of the component)
+ * - [1]: component+variant skips (match a specific variant, keyed as "component\0variant")
+ */
+export type SkipSet = readonly [componentOnly: Set<string>, componentVariant: Set<string>];
+
+function isSkipItem(item: unknown): item is SkipItem {
+  if (typeof item !== 'object' || item === null) return false;
+  const record = item as Record<string, unknown>;
```
lencioni, line 12: I can come back after this merges and see if I can find a good way to remove the typecasting here too.

src/network/__tests__/createExtendsReportSnapRequest.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    assert.strictEqual(call.arguments[0]?.path, '/api/snap-requests/extends-report');
+    assert.strictEqual(call.arguments[0]?.method, 'POST');
+  });
+
+  it('sends extendedSnaps, extendsSha, and project in the request body', async () => {
+    await createExtendsReportSnapRequest('baseline-sha', skippedExamples, config);
+
+    const call = makeHappoAPIRequestMock.mock.calls[0];
+    assert.ok(call);
+    const body = call.arguments[0]?.body as {
```
lencioni, line 65: Typecasting in tests is also not very desirable, but it's usually less of an issue than elsewhere.

lencioni:
We need to adjust the estimated snaps count when there are skipped items

src/cli/index.ts
trotzig: I went with --skip

trotzig:
I'm going to continue in follow-ups

PR #463, opened 2026-04-14
Rename storyFile to file; add file-based skip for E2E

trotzig:
## Summary

- Renames the `storyFile` field in `SkipItem` to `file` across the entire codebase
- Extends `--skip` `file` item support to Cypress and Playwright integrations: screenshots taken inside test files matching a `file` item are skipped
- Cypress: `happoGetIntegrationConfig` resolves `file` paths to absolute so the browser can compare against `Cypress.spec.absolute`; new `happoRecordResolvedSkip` task records skipped `{component, variant}` pairs to a temp file
- Playwright: fixture checks `test.info().file` against resolved `file` paths; matching screenshots are recorded to the same temp file
- E2E finalize (`finalizeAll`): reads the nonce-keyed resolved-skip file (`happo-resolved-skip-<nonce>.json` in tmpdir) to replace `file` items with the collected component/variant pairs before sending to the API

## Test plan

- [ ] All unit tests pass (`pnpm test`)
- [ ] TypeScript type-check passes (`pnpm build:types`)
- [ ] Manual Cypress/Playwright test with `--skip '[{"file":"./src/foo.cy.ts"}]'` skips the expected screenshots

### DIFF
--- a/scripts/getSkip.ts
+++ b/scripts/getSkip.ts
@@ -4,7 +4,7 @@
  * are skipped on each day.
  *
  * - One item uses the `component` form to skip a specific variant.
- * - One item uses the `storyFile` form to skip all stories in a file.
+ * - One item uses the `file` form to skip all stories in a file.
  *
  * Usage:
  *   node scripts/getSkip.ts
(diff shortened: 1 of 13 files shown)


trotzig:
Because of how e2e runs are different from Storybook runs, I don't think it's a good idea to support `file` items for Playwright and Cypress. With `file` items we have to resolve the `component` from the test run during runtime so that we can pass that along in the extends-report API call. But if you are skipping e2e tests you also will most likely not run them at all. Which means that we can't resolve component names from the test run during runtime.

Leaving this here for reference but I'll focus on releasing the new `--skip` feature as-is instead.

PR #466, opened 2026-04-14
Remove --skip support from finalize command

trotzig:
## Summary

- Removes `--skip` handling from `finalizeAll` in `src/e2e/wrapper.ts` (strips the skip body field and the now-unused `Example` type)
- Removes the `happo finalize --skip '[...]'` example from the CLI help text
- Removes the two finalize/skip tests from `index.test.ts`

## Test plan

- [ ] Run `pnpm test` — existing finalize tests should still pass
- [ ] Confirm `happo finalize --skip` no longer silently accepts the flag (it will be ignored)

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -924,6 +924,7 @@ describe('main', () => {
         );
       });
 
+
       describe('cancelling the Happo job', () => {
         withOverrides(
           () => process.env,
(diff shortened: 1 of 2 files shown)


trotzig:
I forgot to clean this up in my previous work on `--skip` and `--only`.

trotzig:
This will have to wait a little. I'm going to bring back the old name briefly and release a patch version.

PR #470, opened 2026-04-16
Apply skip/only filters before chunking in register.ts

trotzig:
## Summary

- `happoSkipped` and `happoOnly` were previously checked inside `nextExample` at iteration time, after chunking had already partitioned the full example list
- This caused uneven chunk work distribution when skip/only was active (e.g. only 3 components out of 1000, but chunking still divided the full 1000)
- Moved both filters into `filterExamples` before the chunk step so chunking operates on the already-filtered set

## Test plan

- [ ] Verify that stories in `happo.only` are correctly captured when chunking is enabled
- [ ] Verify that stories in `happo.skip` are excluded before chunk boundaries are calculated
- [ ] Verify normal (no skip/only) chunked runs are unaffected

### DIFF
--- a/src/storybook/browser/register.ts
+++ b/src/storybook/browser/register.ts
@@ -216,6 +216,23 @@ function filterExamples(
 ): Array<Example> {
   const { chunk, targetName, only } = initConfig;
 
+  if (globalThis.happoOnly) {
+    const happoOnly = globalThis.happoOnly;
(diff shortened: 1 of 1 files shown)


trotzig:
The base reports for the skip and only reports have some bad data in them (leftovers from a bug). I'm going to fix that on the main branch (generate full reports). 

Other than that, this change is working well.

trotzig:
I've re-generated baseline reports on main now.

PR #480, opened 2026-04-22
Guard e2e finalizeHappoReport against same-SHA comparison

lencioni:
## Summary

Follow-up to #479. After that fix, `beforeSha` equals `afterSha` on default-branch builds, but `finalizeHappoReport` in `src/e2e/wrapper.ts` was calling `createAsyncComparison` without the same guard — causing Cypress and Playwright jobs on main to throw:

```
Cannot create an async comparison between the same SHA (beforeSha=..., afterSha=...)
```

The fix applies the existing `beforeSha !== afterSha` guard (already present in `finalizeAll` and `cli/index.ts`) to the `finalizeHappoReport` path as well.

## Test plan

- [ ] Verify Cypress and Playwright jobs on the main branch no longer throw a same-SHA comparison error

### DIFF
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -13,6 +13,7 @@ const config: Config = defineConfig(
   {
     ignores: [
       '**/test-assets/**',
+      '**/__tests__/fixtures/**',
       '.out/**',
       '.happo-out/**',
       'coverage/**',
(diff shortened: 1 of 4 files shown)


trotzig (approved):
Test?

PR #491, opened 2026-05-06
Skip rendering for empty --only and extend full report from baseline

trotzig:
## Summary
- When `--only` is given an empty JSON array (`'[]'`), skip the upload and render pipeline entirely and produce a report whose snaps are all borrowed from the baseline via an extends-report snap request.
- For Storybook, the build still runs so we can enumerate every component and pass the full list as `extendedSnaps`.
- When no baseline can be found, fall back to a normal full run with a clear log message (parallels the existing `--skip` fallback).

## Test plan
- [x] `pnpm test` (379 tests pass; new test added in `src/storybook/__tests__/index.test.ts` for the empty-`--only` case)
- [x] `pnpm build:types`
- [x] `pnpm lint`
- [ ] Manual: run `happo --only '[]'` against a project with a baseline and verify the resulting report extends the baseline with no fresh renders
- [ ] Manual: run `happo --only '[]'` against a project with no baseline and confirm the full-run fallback message and behavior

### DIFF
--- a/src/network/prepareSnapRequests.ts
+++ b/src/network/prepareSnapRequests.ts
@@ -175,11 +175,15 @@ export default async function prepareSnapRequests(
       throw new Error('--only is not supported for the pages integration');
     }
     const { resolvedSkip } = await buildPackage(config, logger, skip, only);
-    const result: PrepareSnapRequestsResult = { snapRequestIds: [] };
-    if (resolvedSkip !== undefined) {
-      result.resolvedSkip = resolvedSkip;
+    if (!resolvedSkip || resolvedSkip.length === 0) {
+      // Without a component list we can't construct a meaningful
+      // extends-report, and the alternative would be an async report with
+      // zero snap requests. Surface the failure instead.
+      throw new Error(
+        'Empty --only run could not enumerate any components from the build. An empty --only run requires a readable Storybook index.json (or equivalent) to know which components to borrow from the baseline.',
+      );
     }
-    return result;
+    return { snapRequestIds: [], resolvedSkip };
   }
 
   const prepareResult =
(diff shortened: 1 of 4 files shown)


src/storybook/__tests__/index.test.ts
```diff
@@ -69,5 +69,18 @@ describe('happoStorybookPlugin', () => {
       });
       assert.strictEqual(result.estimatedSnapsCount, 20);
     });
+
+    it('treats an empty --only array as "borrow everything from baseline"', async () => {
```
lencioni, line 74: Should we add a test that verifies that only with an empty array will still end up showing things that were deleted?

trotzig: Addressing this now.

src/network/prepareSnapRequests.ts
```diff
@@ -165,6 +165,23 @@ export default async function prepareSnapRequests(
...
+  // execution.
+  if (only !== undefined && only.length === 0) {
+    if (config.integration.type === 'pages') {
+      throw new Error('--only is not supported for the pages integration');
+    }
+    const { resolvedSkip } = await buildPackage(config, logger, skip, only);
+    const result: PrepareSnapRequestsResult = { snapRequestIds: [] };
+    if (resolvedSkip !== undefined) {
+      result.resolvedSkip = resolvedSkip;
+    }
```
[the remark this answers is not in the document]
trotzig: Good catch. Took option (b) in b5fcc19: when `buildPackage` returns no components (e.g. unreadable `index.json`), the empty-`--only` path now throws a descriptive error instead of producing an async report with zero snap requests.

PR #580, opened 2026-09-01
Package assets as zstd-compressed tar

lencioni:
Builds asset packages as a tar compressed in a single zstd frame instead of a zip, falling back to zip where zstd isn't available.

> [!IMPORTANT]
> This needs the matching server-side support to be fully deployed before it is released. Until then, Happo workers cannot read the packages this produces.

## Why

Zip compresses each entry independently, so it can never exploit redundancy between the many near-identical files a static package usually holds. Compressing a tar as a single zstd frame does:

| corpus | zip (today) | tar.zst L9 | change |
|---|---|---|---|
| asset-heavy package (101 files, 29 MB raw) | 11.18 MB, 457 ms | 8.68 MB, 210 ms | **−22% size, 2.2x faster** |
| 712 small text files | 0.64 MB, 52 ms | 0.35 MB, 31 ms | **−46% size, 1.7x faster** |

Smaller *and* faster to build. Workers decompress them roughly 7x faster on top of that.

## Fallback

zstd landed in Node's `zlib` in 22.15.0 and 23.8.0. We feature-detect (`typeof zlib.zstdCompressSync === 'function'`) rather than check a version, and produce zip when it isn't there — plenty of CI setups still run older Node.

`HAPPO_ARCHIVE_FORMAT=zip` forces the old format, so anyone hitting a problem with zstd packages can work around it without downgrading.

## Tar writing

Tar is written with `modern-tar`, which has no dependencies of its own and adds ~120 KB. I first wrote this by hand to avoid a dependency at all; the review question "are there lurking bugs a package has already solved?" turned out to be well founded, and it found two:

**A silent header corruption.** The size field is 12 bytes — 11 octal digits plus a NUL — so it tops out at 8 GiB. The encoder padded to 11 digits but never checked the result fit, and `Buffer#write` writes as much as it can, so a larger size produced a 13-byte string that ran past the field and overwrote the start of `mtime`. No error, just a corrupt header. Nothing could reach it today (`validateArchive` caps packages at 60 MB), but the only thing standing between us and it was a check in a different layer.

**A rejection zip never made.** The writer refused any path over 255 *bytes*, while the worker rejects on 255 *UTF-16 code units*. A 200-character CJK path is 600 bytes — well inside what the worker accepts, but the writer threw and failed the entire run, where the same package built as a zip would have worked fine. `modern-tar` encodes long paths with the ustar prefix field where it fits and PAX otherwise, so nothing we can build is rejected now.

For the record, `tar-stream` was the other candidate and is a much bigger ask for a published package: 12 transitive packages and 5.9 MB, including the `bare-*` runtime shims.

## Determinism

The archive hash is what dedupes uploads across machines and CI runs, so identical content has to produce identical bytes everywhere.

- **The library does not give us this for free.** `modern-tar` defaults `mtime` to the current time, which would quietly make every archive unique. Every variable header field — `mtime`, `mode`, `uid`, `gid` — is pinned explicitly. That was never the boilerplate; it's the requirement.
- Compression is **one-shot**. Streaming zstd produces different bytes depending on how the input happens to be chunked; I measured three different digests for the same input at three chunk sizes.
- **Level 9 is deliberate.** The size curve flattens above it (L12 buys ~1% for 2x the CPU), and it is well clear of the fast levels. zstd makes no promise of byte-identical output across library versions, and level 3 output changed between zstd 1.5.6 (Node 22.15) and 1.5.7 (Node 22.18+) while L9 and L19 stayed identical across every Node version we support.

Golden-hash tests pin the exact output for both formats. If a future Node upgrade moves the dedupe key, that shows up as a test failure rather than as a silent re-upload of every package. **If that test ever fires it's a decision to make, not a number to update** — the comment in the test says so.

Output is byte-identical to the hand-written writer for paths that fit the 100-byte name field, so existing packages keep their hashes and are not re-uploaded. Only archives containing a long path change, and those come out a block smaller — PAX beats the GNU long-name record we emitted before.

Worth noting for later: hashing a content manifest instead of the compressed bytes would remove the cross-version risk entirely and let us retune compression without busting the cache. That's a bigger change (client and server semantics together, and it invalidates every existing entry once), so it isn't here.

## Upload changes

Uploads now tell the API which format they are and take the upload `Content-Type` from the signed-url response instead of assuming zip. An API that predates this support ignores the unknown query parameter and returns no content type, so the client falls back to `application/zip` and keeps working — covered by a test.

## Note on size limits

`validateArchive`'s 30/60 MB thresholds measure the compressed archive, so they are now effectively ~2x more permissive on content. That seems right given the limit is about upload and download cost, but it is a behavior change and I left the thresholds alone — happy to retune.

## Testing

- `src/utils/__tests__/createTar.test.ts` — 11 tests, every round trip verified by **extracting with the system `tar`** rather than with a reader of our own: long names, a single path component over 100 bytes, non-ASCII names past 255 bytes, empty files, sizes around the 512-byte block boundary, binary content
- `src/utils/__tests__/deterministicArchive.test.ts` — the full structural suite against **both** formats (29 tests), plus format selection and the golden hashes
- `src/network/__tests__/uploadAssets.test.ts` — the format query parameter, the content type handshake, and the old-server fallback

Full suite: 454 passing.

Also verified end-to-end against the real worker-side reader — 101/101 and 712/712 files byte-identical after a round trip, plus a corpus of PAX-encoded long paths.

### DIFF
--- a/src/utils/__tests__/createTar.test.ts
+++ b/src/utils/__tests__/createTar.test.ts
@@ -17,13 +17,19 @@ function u8(str: string): Uint8Array {
 }
 
 /**
- * Writes a tar to disk and extracts it with the system `tar`, so we are
- * checking our output against a real implementation rather than against our
- * own reader.
+ * Writes a tar to disk and lists it with the system `tar`, returning the entry
+ * names it reports.
+ *
+ * Use this rather than `extractWithSystemTar` whenever a name might not be
+ * representable on the filesystem running the tests — Linux caps a single path
+ * component at 255 bytes and Windows normalizes Unicode — so that we are
+ * testing the archive we wrote rather than the filesystem underneath it.
  */
-async function extractWithSystemTar(
-  tar: Buffer,
-): Promise<Map<string, Buffer>> {
+/**
+ * Writes a tar to disk and extracts it with the system `tar`, so we check our
+ * output against a real implementation rather than against ourselves.
+ */
+async function extractWithSystemTar(tar: Buffer): Promise<Map<string, Buffer>> {
   const dir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'happo-tar-'));
   try {
     const tarPath = path.join(dir, 'archive.tar');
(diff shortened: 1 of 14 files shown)


trotzig (approved):

src/utils/createTar.ts
```diff
@@ -0,0 +1,53 @@
...
+        name: entry.name,
+        size: entry.data.byteLength,
+        type: 'file',
+        ...FIXED_HEADER,
+      },
+      body: entry.data,
+    })),
+  );
+
+  return Buffer.from(packed) as Buffer<ArrayBuffer>;
```
lencioni: Is it possible to avoid type casting here?

src/utils/deterministicArchive.ts
```diff
@@ -133,6 +184,66 @@ async function contentToUint8Array(
...
+async function createZstdArchive(
+  entryDataList: Array<EntryData>,
+): Promise<Buffer<ArrayBuffer>> {
+  const tar = await createTar(entryDataList);
+
+  return zlib.zstdCompressSync(tar, {
+    params: {
+      [zlib.constants.ZSTD_c_compressionLevel]: ZSTD_COMPRESSION_LEVEL,
+    },
+  }) as Buffer<ArrayBuffer>;
```
lencioni: Can we avoid type casting?

src/utils/createTar.ts
lencioni: Good catch — it wasn't needed. `Buffer.from` over a Uint8Array already returns `Buffer<ArrayBuffer>` in the current @types/node. Removed, and verified with a forced clean build (`tsc --build --force`) rather than the incremental one, in case it was being cached.

src/utils/deterministicArchive.ts
lencioni: Same here — `zlib.zstdCompressSync` already returns `Buffer<ArrayBuffer>`. Both casts are gone and the clean build passes.

src/utils/__tests__/createTar.test.ts
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * Use this rather than `extractWithSystemTar` whenever a name might not be
+ * representable on the filesystem running the tests — Linux caps a single path
+ * component at 255 bytes and Windows normalizes Unicode — so that we are
+ * testing the archive we wrote rather than the filesystem underneath it.
+ */
+/**
+ * Writes a tar to disk and extracts it with the system `tar`, so we check our
+ * output against a real implementation rather than against ourselves.
+ */
```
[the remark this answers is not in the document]
lencioni: You're right, that was leftover. I had a `listWithSystemTar` helper that used `tar -tf`, removed it when Windows turned out to mangle non-ASCII in listings, and left its doc comment behind. Merged into one accurate comment that also records when the helper *isn't* usable — long or non-ASCII names that the host filesystem can't represent.

PR #601, opened 2026-09-14
Warn when a Storybook is built in development mode

lencioni:
## Problem

Storybook's `features.developmentModeForBuild` flag defines `process.env.NODE_ENV` as `"development"` in a *production* build, so the built Storybook ships the **development build of the framework**.

From `@storybook/builder-vite` (10.5.10) — the only place in Storybook that sets this in a build:

```js
options.features?.developmentModeForBuild && finalConfig.plugins?.push({
  name: "storybook:define-env",
  config: () => ({ define: { "process.env.NODE_ENV": JSON.stringify("development") } })
});
```

Measured by building this repo's own test Storybook both ways:

| build | React DevTools banner | `react-18` chunk |
|---|---|---|
| default | absent → production | 181,751 B |
| `developmentModeForBuild: true` | **present** | 355,046 B (~2×) |

The size delta understates it — development React's real cost is reconciliation and warning generation, not bytes. Happo renders every story, so this lands on every snapshot in the report.

The flag is opt-in and documented as a debugging aid ("better testability and debuggability"), so a project can easily have it on without anyone connecting it to how long their Happo jobs take. Investigating one slow project recently, establishing this was the cause meant reading 68k lines of worker logs and noticing a React DevTools banner repeating once per story.

## Change

Warn at build time.

The flag is read out of Storybook's own **`project.json`**, which records the *resolved* configuration:

```json
{"features":{"developmentModeForBuild":true}, "framework":{"name":"@storybook/react-vite"}, ...}
```

That's the right source: it works whatever the framework and however the flag got set, where parsing an arbitrary TypeScript `main.ts` would not. And the file is already in hand — we delete it after every build so it doesn't get uploaded. The check runs immediately before that delete.

Prebuilt packages (`usePrebuiltPackage`) get the same check, since the flag costs the same whoever ran the build. Their `project.json` is left in place — a package we didn't build isn't ours to tidy up.

A missing, unreadable, or malformed `project.json` is silently ignored; there'd be nothing to warn about and nothing actionable in the error.

## Output

```
[HAPPO] This Storybook was built with `features.developmentModeForBuild` enabled, so it ships the development build of your framework. That is substantially slower to render, and Happo renders every story — expect slower jobs and a higher chance of timeouts. Remove the flag from your `.storybook/main` config to build in production mode.
```

## Testing

- New `warnIfDevelopmentModeBuild.test.ts` — 5 tests: flag on, flag absent, flag explicitly false, no `project.json`, malformed `project.json`
- Full storybook suite: 42 tests pass
- `pnpm lint` and `pnpm build:types` clean
- Verified end-to-end against two real Storybook builds produced from `src/storybook/__tests__/storybook-app` — warns on the dev-mode one, silent on the default one


trotzig (approved):
I wonder if there is a way we could expose this information on the happo.io side of things as well?

PR #602, opened 2026-09-14
Build Storybook previews without the manager UI

lencioni:
Rebased onto `main` (conflicted with #601, which landed in the same part of `src/storybook/index.ts`), and **`previewOnly` now defaults to on** per the discussion with @trotzig.

## What

Storybook builds now pass `--preview-only`, leaving out the manager UI. `previewOnly: false` opts back in.

```ts
integration: { type: 'storybook', previewOnly: false }  // keep the manager
```

## Why

Happo only ever loads `iframe.html`. The manager — the Storybook sidebar app — is weight nobody benefits from in an uploaded package. Measured on this repo's own test Storybook:

| build | size | files |
|---|---|---|
| with manager | 8,560 KB | 30 |
| **preview only (new default)** | **1,808 KB** | 23 |

**79% smaller** — less to upload from CI, less to transfer to a worker, less to unpack there.

It drops only `index.html`, `sb-manager/*` and `sb-addons/*`. `iframe.html`, `index.json`, the preview bundle, story chunks, fonts and static dirs all remain. I verified rendering works rather than trusting the file list: served a preview-only build and loaded a story in a browser — it rendered, with Happo's preview decorator live (its own `Ignoring forceHappoScreenshot...` message in the console). The decorator comes from the preset's `config()`, which is preview-side; only the manager addon is dropped.

## Why there's still an opt-out

**The manager is not dead weight for people.** A preview-only package isn't a browsable Storybook — no `index.html`, no sidebar — so anyone who downloads a built package to debug it has to reach stories by visiting `iframe.html?id=<storyId>&viewMode=story` by hand. @lencioni raised this, which is why the escape hatch exists.

## Version gate

`--preview-only` landed in **Storybook v9** and doesn't exist in v8, which we still support. Verified against all three CLIs (`storybook@8`, `@9`, `@10`) — v8 has no such flag and fails the build outright if handed it.

On v8 the flag is dropped and the build carries on. That's **silent on the default** (nobody has done anything to be told about) and warns only when `previewOnly: true` was asked for explicitly:

```
[HAPPO] Ignoring `previewOnly` because it needs Storybook v9 or later.
```

The option only ever makes the package smaller, so failing a build over it would trade a working report for an optimization.

## Testing

Rewritten around the new default, running real builds:

- **leaves out the manager UI by default** — no `index.html`, `sb-manager`, `sb-addons`
- **still builds everything Happo renders from** — `iframe.html`, `index.json`, preview assets present. This is the one that matters: if `--preview-only` ever starts dropping one of those, rendering breaks rather than just getting smaller.
- **keeps the manager UI when `previewOnly: false`** — the opt-out actually opts out
- **produces a smaller package than one built with the manager**

Full storybook suite: 45 pass, 0 fail, 1 skipped. `pnpm lint` and `pnpm build:types` clean.

One gap worth naming: the repo only has Storybook v10 installed, so `index-v8.test.ts` exercises a v8-*compatible app config*, not a real v8 CLI. The version gate itself I checked by hand against `npx storybook@8`.


lencioni:
I'm not sure that this is actually useful or good.

trotzig (approved):
This seems like a good idea in my opinion! I don't think we encourage people to download the built assets and look at the preview anyway, so I actually think this could be turned on by default.

lencioni:
Ok! I'll rebase it and get it ready for being on by default then.

lencioni:
For storybook v10 in this repo, this change reduces the size of the asset package (compressed with zstd) from 1.4 MiB to 554 KiB. Not bad!

PR #605, opened 2026-09-15
Add an allowedHostnames target option

lencioni:
Client half of an opt-in network allowlist for renders. The worker half is happo/happo-snap#1347, which is where the interesting parts are.

## What it does

Names the hostnames a target expects the browser to reach while rendering; the worker refuses everything else before it leaves the browser.

```js
targets: {
  chrome: new RemoteBrowserTarget('chrome', {
    viewport: '1024x768',
    allowedHostnames: ['fonts.gstatic.com', '*.mycdn.example.com'],
  }),
}
```

A snapshot that loads a font, a script or an image from a third party changes when that third party does, and fails when it's down — so the fewer hostnames a target needs, the more reproducible it is.

Entries are hostnames matched exactly; `*.example.com` covers subdomains but not the apex, like a wildcard cert. Ports and paths are ignored, so `https://example.com/assets` and `example.com` mean the same thing. Leave it unset (the default) and nothing is blocked.

You don't have to guess what to put in the list. The worker logs the external hostnames a run reached for **whether or not a list is set**, so run once without one and read them off the snap-request's logs:

```
External requests: 12 to 2 hostnames: fonts.gstatic.com (x11), cdn.example.com
```

Once a list is in force the same line splits into allowed and blocked.

## Why this is mostly a type

The option rides to the worker in `otherOptions`, like every other target option the client has no opinion about, so there's no plumbing to add — just the type, its documentation, and `TargetWithDefaults` so the resolved target carries it too.

It's still worth pinning with tests. An allowlist that silently fails to arrive looks exactly like one that allowed everything, which is the worst failure mode a feature like this has. And `[]` has to survive the trip intact, because an empty list and an absent one mean opposite things — block everything external, versus block nothing. That distinction is what step 2 turns on.

## Reviewer notes

- Marked `@experimental`. The intended follow-up is to default this to `[]`, blocking all external requests unless a target opts out. That's a breaking change for anyone whose stories load third-party assets, which is why this lands first and opt-in — targets can find out what they'd need, and set it, before the default moves.
- **Not supported on `ios-safari` / `ipad-safari`**, where the browser can't be pointed at the proxy that does the blocking. Those targets log that the option had no effect rather than leaving you to assume it did. It's typed on `DesktopTarget` only, so the compiler catches it too.
- **This is useful only with a worker that understands it.** Against an older worker the option is accepted and silently ignored — the same shape as every other `otherOptions` addition, but worth knowing given what this one claims to do.

## Testing

`pnpm lint`, `pnpm build:types` and `pnpm test` (475 passing) all clean.


trotzig (approved):
In merrykat this setting ended up in admin because of security concerns. I don't think that we need them in admin for Happo, I think it's fine if they live here. The setting is more about stability and less about security.

lencioni:
The workers have been updated, so I think we can land this now.

PR #610, opened 2026-09-19
Make the Storybook integration work without a preview.js import

lencioni:
## Why

A new account signed up yesterday. Their first Happo job failed with `Timed out while waiting for window.happo`. Their second, five minutes later, rendered every story as the literal text `[object Object]`. Their third worked.

Both failures are ours to prevent.

**Job 1** — the Storybook integration required `import 'happo/storybook/register'` in `.storybook/preview`, and nothing catches a missing one until a worker has been queued, has polled for ten seconds, and has given up.

**Job 2** — reproduced from their uploaded package. They had followed the setup and added `happoDecorator`, which wrapped every story in a React element via `createElement`. Their Storybook is `@storybook/web-components`, so that element reached a Lit template, which stringified it. Opening their package in a plain browser shows `[object Object]` on every story, with no Happo involved at all.

## What changed

**Ship the client runtime inside the package.** `browser/register.ts` reaches Storybook entirely through globals — every one of its Storybook imports is `import type` — so it builds as a standalone IIFE that `buildStorybookPackage()` copies in and links from the head of `iframe.html`. The preview.js import is no longer needed for a run to work. It stays supported and documented, for the helpers it exports.

**Share the runtime's state through `globalThis`.** This one is subtle and worth a look. A project that imports anything from `happo/storybook/register` — `forceHappoScreenshot` in a story file, say — gets a second copy of the module bundled into the preview, and Storybook loads story files lazily. That copy can therefore be evaluated *during* a run: `init()` awaits the story store, the store loads the story files, and the second copy reinstalls `window.happo` with an empty `examples`. The next `nextExample()` call then throws `Missing examples`. Moving the mutable state onto `globalThis` means both copies operate on the same run. I hit this while verifying the change end to end, against the repo's own test app.

**Make the addon decorator renderer-agnostic.** It now uses Storybook's `useEffect` — which works under every renderer — and returns the story untouched instead of wrapping it in a React element.

**Check the built package before uploading it.** A package with nowhere to put the runtime, or with no stories in it at all, now fails on the developer's terminal rather than coming back from a worker minutes later as a timeout or an empty report.

## Verification

The no-import path is now exercised in CI, which it was not before — every fixture imported the runtime, so every job would have passed just as happily without the injection.

`storybook-app-no-register` is a fixture that imports the runtime nowhere. `storybook-app` could not serve that purpose — one of its stories imports `forceHappoScreenshot`, so the runtime is in its bundle whatever its preview does — and hanging it off `storybook-app-v8` would have tied it to a fixture #611 renames and a Happo job #611 drops.

`injectedRuntime.spec.ts` builds that fixture's package, serves it, and drives `window.happo.init()` / `nextExample()` the way a worker does, against whichever Storybook version is installed. It also asserts the package bundles no second copy of the runtime, so an import added to the fixture fails there rather than quietly removing the coverage. Both checked by breaking them on purpose: neutering the injected runtime fails the `window.happo` wait — the customer-facing symptom — and adding an import fails the second check by name.

A `storybook-no-register` job renders the same fixture on real workers, so the path is covered both ways: the spec is the fast signal, the job is the one a customer would notice first. Nine snapshots a run — three examples across the three targets in the base config. The first run creates the project and has no baseline, so its comparison will be all new.

Manually, before those existed: built the v10 test Storybook with the import removed and drove it to exhaustion — 25 examples, including the two `forceHappoScreenshot` steps that come from the lazily-loaded second copy. Restored the import and re-ran: 26 examples, with `setThemeSwitcher` from the preview copy taking effect.

## Note for review

`docs` has a companion PR. The docs there say "recent versions" rather than naming one — worth pinning to the real number once this is released.


trotzig (approved):

src/storybook/__tests__/storybook-app-v8/preview.ts
```diff
@@ -1,7 +1,16 @@
-import '../../browser/register.ts';
```
lencioni, line 1: It looks like storybook v8 is being removed in the next breaking change https://github.com/happo/happo/pull/611

We will need to make sure that we are still testing storybook without the preview.js import in that PR.

PR #611, opened 2026-09-20
Breaking changes for the next major release

trotzig:
Collects the breaking changes for the upcoming major release. Each is its own commit, so anything that doesn't belong can be dropped without unpicking the rest.

## Config and CLI removals

| Removed | Replacement |
| --- | --- |
| `target.chunks` | Happo chunks for you, from the size of the run |
| `useFullPageFallbackForTallScreenshots` | Nothing — the option no longer exists |
| `githubApiUrl` | Nothing — statuses are posted server-side |
| `--githubToken` | Nothing — statuses are posted server-side |
| `--skippedExamples` | `--skip` |

A config that still sets a removed **config** option now fails with an error naming the option and saying what to do instead, rather than having it silently ignored. The tables that drive this live in `src/config/loadConfig.ts`, so the next removal is a one-line addition. Removed **CLI flags** are rejected by `parseArgs` as unknown options.

`useFullPageFallbackForTallScreenshots` never existed in this package — it is a happo.io v5 leftover — so there was nothing to delete. It is wired into the migration errors so configs carried over from v5 get told rather than ignored.

## Behavior changes

**`allowedHostnames` defaults to `[]`**, so a target blocks every external request unless it names the hostnames it needs. A snapshot that loads a font, a script or an image from somewhere else changes when that somewhere else does and fails when it is down, so reaching outside the package is now opt-in. Run once and read the blocked hostnames off the snap-request's logs to build the list. Mobile Safari targets are left without a list: the worker can't point those browsers at the proxy that does the blocking and reports that the option had no effect, so defaulting them to `[]` would put that line in every run for nothing. The three targets in `happoconfigs/happo.config.ts` already set `[]` by hand and now get it from the default.

**Chunking is always automatic**: ~100 snapshots per chunk, capped at 20. This already happened for static packages that reported an `estimatedSnapsCount`. It now also covers the `pages` integration (chunked on page count) and e2e runs (chunked on snap-payload count), **both of which used to be a single chunk unless you set `chunks` yourself**. A 250-page or 250-snapshot run becomes 3 snap-requests instead of 1. This is the one change nobody opted into — worth a deliberate yes or no.

**`integration.autoApplyPseudoStateAttributes` defaults to `true`** for Cypress and Playwright, and in `takeDOMSnapshot` itself. Set it to `false` to keep managing the `data-happo-*` attributes by hand. The explicit `true` came out of the two configs in `happoconfigs/` so the integration suites exercise the default path.

**A custom integration's `build()` must return `estimatedSnapsCount`.** With `chunks` gone it is the only thing that tells Happo how many workers to use, and only `build()` knows the number — leaving it optional would have quietly dropped anyone who used `chunks` to a single worker. Required in the type and at runtime, with an error that says what to return and why. A value that isn't a non-negative finite number is rejected the same way. The repo's own custom config was an instance of the problem and now declares its count.

**Client-side PR comment posting is gone entirely** (`postGitHubComment`, `--githubToken`, `environment.githubToken`). The server posts statuses itself now, GitHub Enterprise included, so the status-image comment fallback has nothing left to do. `statusImageUrl` also comes off the created-async-comparison result: nothing reads it any more, and requiring it could only turn a working run into a failure.

## Platform and dependency floors

- **Node**: `engines` drops the `^23.6.0` range. Node 23 went end-of-life in June 2025. Now `^22.18.0 || >=24.0.0`.
- **Storybook**: floor raised from v8 to v9. This removes the duplicated `addon-v8` bundle and its manager-api compat shim, the `MIN_PREVIEW_ONLY_VERSION` gate (every supported version has `--preview-only`), the v8 happo config and the v8 CI job. The v8 fixture app survives as `storybook-app-minimal` — it still covers a config with no addons beyond Happo's own preset, so that coverage isn't lost.

## Internal cleanup

`RemoteBrowserTarget` no longer falls back to one `POST /api/snap-requests` per chunk when `/api/snap-requests/bulk` answers 404 or 501. happo.io is hosted, so no deployment is old enough to need it; a bulk endpoint that is missing now surfaces as an error. Per-item retry of individual bulk failures stays. `MAX_BULK_ITEMS_PER_REQUEST` batching went with it — with chunks capped at 20 it could never trigger.

Also widened `.out-*` in `.gitignore` and `eslint.config.ts`, which both had `.out-v8` hardcoded.

## For reviewers

- `allowedHostnames: []` as a default leaves no documented way to say "allow everything" — entries are exact hostnames or `*.` subdomain prefixes, and there is no wildcard in the client. If the worker treats something like `['*']` as allow-all, that is worth documenting as the escape hatch for anyone who can't enumerate their hostnames; if it doesn't, worth deciding whether it should.
- The auto-chunking of `pages` and e2e runs is the judgement call in here. If you'd rather keep those single-chunk and auto-chunk static packages only, it's a two-line change in `src/config/RemoteBrowserTarget.ts`.
- Removing Storybook v8 drops the `storybook-v8` entry from the happo job orchestration list in CI.
- `tsc --build --force`, `eslint .`, the dist build and the unit suite (474 passing, 1 pre-existing skip) are all green. Prettier reports the same set of already-unformatted files as `main` — no new drift.
- The integration suites that hit the real Happo API (`test:storybook`, `test:cypress`, `test:playwright`, `test:custom`, `test:pages`) haven't been run locally; CI covers them.


lencioni:
We should also flip allowedHostnames default value to `[]` in the next breaking change.

Also, we should look at these and see if there's anything we need to do in a minor release up front that can help people update to the new major release more incrementally. i.e. can people adopt the breaking changes or set their preferred configuration one by one before they update? And we should add deprecation notices in a non-major release ahead of any of these wherever possible, to help people adjust.

.github/workflows/test.yaml
```diff
@@ -75,15 +75,6 @@ jobs:
       - uses: ./.github/actions/setup-env
       - run: pnpm test:pages
 
-  storybook-v8:
```
lencioni, line 78: ~Just noting that the v8 storybook is going to be testing storybook without the preview.js import and when we remove it we'll want to make sure that is tested end to end somewhere. https://github.com/happo/happo/pull/610#discussion_r4062520619~ Update: looks like Claude saw my comment and added a separate build for this. The data point below about v8 still stands, however.

But also, storybook v8 is the second most popular version still, so we might want to wait a bit longer before we drop support in this package:

Image

PR #612, opened 2026-09-21
Say why the Storybook build failed

lencioni:
## What's wrong

A customer's job ([#2769133](https://happo.io/a/126/jobs/2769133)) failed with `Failed to build static storybook package`. That message is true of every possible failure, which makes it useless for all of them — and their CI logs aren't something we can see.

The message is a dead end by construction. `buildStorybook()` spawned the build with `stdio: 'inherit'`, so Storybook's own output went straight to the CI log and never reached us, and the error we threw dropped the exit code, the signal, the command, and every word Storybook printed. `formatFailureMessage` had nothing to work with, so the job page got a sentence that can't distinguish an OOM kill from a typo in a story file.

**I can't tell you that customer's root cause, because the client threw it away.** This PR makes the next one answerable.

## What changed

- **Capture the build output.** stdout/stderr are piped and written straight back out, so a terminal or CI log still sees the output live and in full — byte-identical, and `FORCE_COLOR` keeps colors when we're on a TTY. A bounded tail (20 KB) is kept for the error. stdin stays inherited so an `npx` install prompt still works.
- **Put the reason where it survives.** New `formatStorybookBuildFailure` builds the error: exit code, signal, command, and the last 30 lines with ANSI stripped. Happo keeps only the first line of the message and caps it at 120 characters, so the one line most likely to explain the failure goes first — preferring Storybook's own structured `SB_*` error code, falling back to a builder error line, never a stack frame.
- **Name the two failures worth naming.** A SIGKILL or exit 137 is called out as a probable out-of-memory kill with the `--max-old-space-size` fix — that's how a Storybook build usually dies on a CI machine but not on a laptop. A missing `iframe.html` now says where we looked, with different advice depending on whether we ran the build or were handed a prebuilt package.
- **Fix an unrelated crash found next door.** The spawn had no `'error'` listener. I confirmed an unspawnable binary emits an unhandled `'error'` event and takes the process down with an uncaught `ENOENT`, escaping the `catch` in `src/cli/index.ts` that's supposed to cancel the job — so the job hung in the queue with no message at all. Now it rejects with the command and the underlying error.

## Effect on the job page

```
before: Storybook run failed: Failed to build static storybook package.

after:  Storybook run failed: SB_CORE-SERVER_0006 (MainFileMissingError): No
        configuration files have been found in your conf… Review the logs from
        your CI job for the full output: https://ci.example.com/job/1
```

That's from a real failing build, not a fixture.

## Notes for the reviewer

- **`'exit'` → `'close'`** is load-bearing. With pipes, `'exit'` can fire before the streams drain; `'close'` guarantees the captured output is complete before we build the error from it.
- **CI output is unchanged.** `FORCE_COLOR` is only set when our own stdout is a TTY, which it isn't in CI — so CI logs stay exactly as colorless as they were under `inherit`.
- **The tests go through `formatFailureMessage`.** The value of a better error is entirely in what survives truncation, so the job-page assertions exercise the real formatter rather than trusting the raw message.
- `mise exec -- pnpm lint`, `build:types`, and all 509 tests pass, including the real v8 and v10 Storybook builds that exercise the new piped success path. I verified the spawn-crash regression test fails without the fix.

## Still open: what actually broke for that customer

This PR improves the diagnosis, not the cause — I couldn't determine the cause, which is the point. Two leads worth chasing separately:

1. **`--preview-only` is my main suspect for "started failing."** It shipped in v6.17.0 on Sept 14 and is on by default, which fits a project that broke without changing anything. I checked the published tarballs: the flag exists in every 9.x back to 9.0.0 and in 10.x, so the `>= 9` guard is sound. But it's the only recent change that alters the command we hand Storybook. Worth asking the customer for their happo version and whether pinning to v6.16.0 fixes it.
2. **`getStorybookVersionFromPackageJson` has a hole that produces exactly this symptom.** It takes the first digits anywhere in the specifier, so `"catalog:react19"` reads as Storybook **19** and skips the `node_modules` fallback written specifically for catalogs. On a project actually running Storybook 8 that means we pass `--preview-only` to a CLI that rejects it — a hard build failure. Left out of this PR to keep it focused; tracked separately.


src/storybook/index.ts
```diff
@@ -133,15 +136,72 @@ async function buildStorybook({
...
+      capturedOutput += chunk.toString('utf8');
+      if (capturedOutput.length > MAX_CAPTURED_OUTPUT_CHARS) {
+        // Keep the end rather than the start: a build that fails partway
+        // through says why on its way out, not on its way in.
+        capturedOutput = capturedOutput.slice(-MAX_CAPTURED_OUTPUT_CHARS);
+      }
+    }
+
+    spawned.stdout?.on('data', (chunk: Buffer) => capture(chunk, process.stdout));
+    spawned.stderr?.on('data', (chunk: Buffer) => capture(chunk, process.stderr));
```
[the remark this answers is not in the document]
lencioni: Good catch — this was a real regression, not a theoretical one.

`stdio: 'inherit'` never had the problem: the child wrote to the fd directly, so the kernel blocked *it* when the pipe filled. Writing each chunk from the `'data' handler drained the child as fast as it produced and buffered the difference in our process instead.

Fixed in bbb1fbb by forwarding with `pipe()`, which pauses the child's stream when our stdout cannot keep up. Measured against a writable that never completes a write:

- `'data' handler + write()` — allocated 3.3 GB and died with `JavaScript heap out of memory` within 300ms
- `pipe()` — stops at 64 KB, the high-water marks and nothing more

The capture stays a separate `'data' listener rather than a transform. `pipe()` pauses the source under backpressure, which stops those events too, so what we capture stays in step with what has actually been written out. Also passing `end: false` so the child's stdout ending does not close ours for the rest of the run.

src/storybook/formatStorybookBuildFailure.ts
```diff
@@ -0,0 +1,154 @@
...
+// eslint-disable-next-line no-control-regex -- matching ANSI escapes requires them
+const ANSI_PATTERN = /\u001B\[[0-9;]*[A-Za-z]/g;
+
+/**
+ * Storybook colorizes its output and draws progress spinners. Those escape
+ * codes are fine on a terminal but turn into visible junk once the text is
+ * embedded in an error message and forwarded to the Happo job page.
+ */
+export function stripAnsi(value: string): string {
+  return value.replaceAll(ANSI_PATTERN, '');
```
lencioni: This is a node built-in now: util.stripVTControlCharacters

Let's use that instead of rolling our own.

lencioni: Done in b85a545 — much better, thanks. It also resolves the two Copilot findings about the regex, since they were the same bug.

The hand-rolled version only matched CSI sequences made of digits and semicolons ending in a letter, so it missed the private-mode cursor codes a spinner emits, and OSC sequences entirely. Against a realistic spinner frame (cursor hide, color, line clear, cursor show):

```
old regex                -> "ESC[?25l| buildingERR! it broke ESC[?25h"   (escapes survive)
stripVTControlCharacters -> "| buildingERR! it broke"                    (clean)
```

Those would have shown up as visible junk on the job page, which is the thing the stripping exists to prevent.

I dropped the unit tests for the helper — no value in testing a Node built-in. What is worth testing is that the formatter leaves nothing behind, so that case now feeds a full spinner frame through `formatStorybookBuildFailure` and asserts no escape byte survives. I checked it fails against the regex it replaces.

